PR #38, opened 2025-10-06
Remove unnecessary configs and move config helpers

tomaisthorpe:
- Removed environment and Aikido configs to simplify configs, the values are now just passed to the agent
  - Some of this may come back once the agent is merged in
- Move config helpers into config package

I've left the mutex exported for now, as there's a method that updates the config from the agent. We should revisit this once the agent is merged in.


internal/config/config.go
```diff
@@ -50,8 +37,78 @@ type CloudConfigData struct {
...
+}
+
+func GetEndpoints() []aikido_types.Endpoint {
+	CloudConfigMutex.Lock()
+	defer CloudConfigMutex.Unlock()
+
+	return CloudConfig.Endpoints
+}
+
+func keyExists[K comparable, V any](m map[K]V, key K) bool {
```
bitterpanda63, line 111: maybe lets get this back into the helpers package but as a separate file?

internal/config/config.go
```diff
@@ -50,8 +37,78 @@ type CloudConfigData struct {
...
+	defer CloudConfigMutex.Unlock()
+
+	return keyExists(CloudConfig.BypassedIPs, ip)
+}
+
+func GetEndpoints() []aikido_types.Endpoint {
+	CloudConfigMutex.Lock()
+	defer CloudConfigMutex.Unlock()
+
+	return CloudConfig.Endpoints
```
bitterpanda63, line 108: Possible rsik here that endpoints change while thread is reading them? (yes this risk was already there before, just making sure we are still aware)

internal/config/config.go
tomaisthorpe: Personally, I'm not sure this is much of an improvement over just doing:

```go
_, exists := m[k]
return exists
```

I'd prefer to either keep it here, or remove entirely.

internal/config/config.go
tomaisthorpe: The endpoints aren't changed by any part of the code at the moment.
The only thing we have is the agent which replaces the whole slice [here](https://github.com/AikidoSec/firewall-go/blob/8032ea02f6b5e1b8ccfa517e722b8f90955223ec/internal/grpc/config.go#L67).

I think it's another thing we should review again once the agent is merged in.

bitterpanda63: Yes, was talking about an update from the cloud indeed

bitterpanda63 (approved):

PR #44, opened 2025-10-07
feat: use `IsLoopback` for `isLocalhostIP`

tomaisthorpe:
Using `net` approach for detecting localhost IP, should be more reliable.


internal/http/ip_allowed_to_access_route.go
```diff
@@ -37,3 +37,11 @@ func ipAllowedToAccessRoute(ip string, matches []aikido_types.Endpoint) bool {
 func ipAllowed(remoteAddress string, endpoint aikido_types.Endpoint) bool {
 	return slices.Contains(endpoint.AllowedIPAddresses, remoteAddress)
 }
+
+func isLocalhostIP(ip string) bool {
```
bitterpanda63, line 41: we use this in other codebases on multiple places, would still keep in helper?

bitterpanda63 (approved):

PR #48, opened 2025-10-09
Remove context store and use `context.Context`

tomaisthorpe:
Using the standard `context.Context` to propagate context throughout the request flow as this simplifies the implementation a lot. For DB calls, this should work in _most_ cases, but will rely on apps using `QueryContext`, `ExecContext`, etc. It won't work for apps using legacy non-context methods.

Later we will need to re-add the runtime context storage to ensure context is always available, regardless of whether the developers are using the context methods.

Known issue: this will cause the `os` sink to no longer function as they don't accept a context, so we can't guarantee it will be there when we wrap it. That will work again once we re-add the runtime context storage.


internal/http/on_init_request.go
```diff
@@ -12,27 +14,31 @@ type Response struct {
 }
 
 func OnInitRequest(ctx context.Context) *Response {
-	if config.IsIPBypassed(ctx.GetIP()) {
+	httpCtx := request.GetContext(ctx)
```
bitterpanda63: rename to reqCtx

bitterpanda63 (approved):

PR #49, opened 2025-10-09
chore: remove protobuf

tomaisthorpe:
Removing remaining pieces of protobuf


Makefile
```diff
@@ -3,7 +3,6 @@ prepare: check_binaries
 	git submodule update --remote --merge
```
hansott: Not needed anymore?

bitterpanda63: needs to go indeed

bitterpanda63 (approved):

PR #50, opened 2025-10-09
chore: remove grpc package

tomaisthorpe:
Call agent directly instead of intermediate grpc package.
Moved config logic into config package.


internal/config/cloud_config.go
```diff
@@ -64,33 +72,33 @@ func setCloudConfig(cloudConfigFromAgent *aikido_types.CloudConfigData) {
...
-	config.CloudConfig.BlockedIPs = map[string]config.IPBlockList{}
+	CloudConfig.BlockedIPs = map[string]IPBlockList{}
 	for ipBlocklistSource, ipBlocklist := range cloudConfigFromAgent.BlockedIPsList {
-		config.CloudConfig.BlockedIPs[ipBlocklistSource] = buildIPBlocklist(ipBlocklistSource, ipBlocklist.Description, ipBlocklist.Ips)
+		CloudConfig.BlockedIPs[ipBlocklistSource] = buildIPBlocklist(ipBlocklistSource, ipBlocklist.Description, ipBlocklist.Ips)
 	}
 
 	if cloudConfigFromAgent.BlockedUserAgents != "" {
-		config.CloudConfig.BlockedUserAgents, _ = regexp.Compile("(?i)" + cloudConfigFromAgent.BlockedUserAgents)
+		CloudConfig.BlockedUserAgents, _ = regexp.Compile("(?i)" + cloudConfigFromAgent.BlockedUserAgents)
```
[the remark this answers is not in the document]
tomaisthorpe: These are the regex that come directly from Aikido

bitterpanda63 (approved):

agent/agent.go
```diff
@@ -80,6 +80,7 @@ func OnUser(id string, username string, ip string) {
 }
 
 func OnAttackDetected(attack *aikido_types.DetectedAttack) {
+	log.Debugf("Reporting attack")
```
hansott, line 83: forgotten debug statement?

bitterpanda63: makes sense as a debug statement, but think we want an info-level log with all the attack info as well

tomaisthorpe: It was in the original function, but not sure if it's necessary? I can remove it if you think makes sense to.

hansott: yup, we can fix later but indeed, when an attack occurs, this should be logged (we can do that later?)

PR #55, opened 2025-10-13
Move rate limiting types into `ratelimiting` pkg

tomaisthorpe:
- Moved types and globals related to rate limiting into the `ratelimiting` pkg
- Renamed `rate_limiting` pkg -> `ratelimiting`


internal/agent/ratelimiting/rate_limiting.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Mutex is used to sync access across the go routines
+	Mutex sync.RWMutex
+
+	// Channel and Ticker for the rate limiting background routine
+	channel = make(chan struct{})
+	ticker  = time.NewTicker(MinRateLimitingIntervalInMs * time.Millisecond)
+)
+
+func advanceQueuesForMap(config *Config, countsMap map[string]*Counts) {
```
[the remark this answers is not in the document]
tomaisthorpe: Agree with this, I'm going to refactor how rate limiting works, so will address this soon.

internal/agent/cloud/common.go
```diff
@@ -45,54 +45,54 @@ func ResetHeartbeatTicker() {
...
-		k := RateLimitingKey{Method: newEndpointConfig.Method, Route: newEndpointConfig.Route}
+		k := ratelimiting.Key{Method: newEndpointConfig.Method, Route: newEndpointConfig.Route}
 		UpdatedEndpoints[k] = true
 
-		rateLimitingData, exists := globals.RateLimitingMap[k]
+		rateLimitingData, exists := ratelimiting.Map[k]
 		if exists {
 			if rateLimitingData.Config.MaxRequests == newEndpointConfig.RateLimiting.MaxRequests &&
-				rateLimitingData.Config.WindowSizeInMinutes == newEndpointConfig.RateLimiting.WindowSizeInMS*MinRateLimitingIntervalInMs {
+				rateLimitingData.Config.WindowSizeInMinutes == newEndpointConfig.RateLimiting.WindowSizeInMS/ratelimiting.MinRateLimitingIntervalInMs {
```
bitterpanda63: mm suddenly switched to division? should be multiplication still

internal/agent/cloud/common.go
```diff
@@ -45,54 +45,54 @@ func ResetHeartbeatTicker() {
...
 		}
 
 		log.Infof("Got new rate limiting endpoint config and storing to map: %v", newEndpointConfig)
-		globals.RateLimitingMap[k] = &RateLimitingValue{
-			Config: RateLimitingConfig{
+		ratelimiting.Map[k] = &ratelimiting.Value{
+			Config: ratelimiting.Config{
 				MaxRequests:         newEndpointConfig.RateLimiting.MaxRequests,
-				WindowSizeInMinutes: newEndpointConfig.RateLimiting.WindowSizeInMS / MinRateLimitingIntervalInMs,
+				WindowSizeInMinutes: newEndpointConfig.RateLimiting.WindowSizeInMS / ratelimiting.MinRateLimitingIntervalInMs,
```
bitterpanda63: this code doesnt make sense, its just a division by 60 * 1000, but has nothing to do with this "MinRateLimitingIntervalInMs"

internal/agent/cloud/common.go
bitterpanda63: on further review, division is correct, but I don't think we should be using a vaguely named constant here, makes it confusing

internal/agent/ratelimiting/rate_limiting.go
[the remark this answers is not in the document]
bitterpanda63: can take a look here at node.js or other agents, most have a similar sliding window rate limiter

internal/agent/cloud/common.go
tomaisthorpe: Added method to convert between ms and minutes, constant was misused.

bitterpanda63 (approved):

PR #58, opened 2025-10-13
Add simple tests for gin/echo sources

tomaisthorpe:
- Added some simple tests for gin/echo sources to check context is set
- Add build step in CI for sample apps
- Updated Query field in context to `url.Values`  instead of map


internal/request/context.go
```diff
@@ -2,13 +2,14 @@ package request
...
 	"encoding/json"
+	"net/url"
 	"sync"
 )
 
 type Context struct {
 	URL                string
 	Method             *string
-	Query              map[string][]string
+	Query              url.Values
```
bitterpanda63: mm actually disagree here, assuming they are the same type right? I think the original code makes it clearer how the values are stored, since we are going to extract them that's more important imo.

tomaisthorpe: I don't think `map[string][]string` makes it clearer, I think for these structs the semantic reasoning is more important than the underlying storage. But, I'll split out into a separate PR!

bitterpanda63: mm ok, can be topic of different pr idd

bitterpanda63 (approved):

PR #62, opened 2025-10-13
feat: pass config structs to agent instead of json

tomaisthorpe:
Left over from removing gRPC, passing config structs directly instead of using JSON.
Removed socket path config as no longer used.

### DIFF
--- a/internal/agent/globals/globals.go
+++ b/internal/agent/globals/globals.go
@@ -7,10 +7,10 @@ import (
 )
 
 // Local config that contains info about socket path, platform, library version...
-var EnvironmentConfig EnvironmentConfigData
+var EnvironmentConfig *EnvironmentConfigData
 
 // Aikido config that contains info about endpoint, log_level, token, ...
-var AikidoConfig AikidoConfigData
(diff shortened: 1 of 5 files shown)


internal/agent/globals/globals.go
```diff
@@ -7,10 +7,10 @@ import (
 )
 
 // Local config that contains info about socket path, platform, library version...
-var EnvironmentConfig aikido_types.EnvironmentConfigData
+var EnvironmentConfig *aikido_types.EnvironmentConfigData
 
 // Aikido config that contains info about endpoint, log_level, token, ...
-var AikidoConfig aikido_types.AikidoConfigData
+var AikidoConfig *aikido_types.AikidoConfigData
```
bitterpanda63, line 13: assuming we use pointers now since the data is already in the same process

bitterpanda63 (approved):

internal/agent/globals/globals.go
tomaisthorpe: yep, but there's also a mutex in there which doesn't work when you're copying by value

PR #68, opened 2025-10-14
Remove global cloud config storage

tomaisthorpe:


internal/vulnerabilities/attack.go
```diff
@@ -79,7 +80,7 @@ func GetAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types
 			Kind:      string(res.Kind),
 			Operation: res.Operation,
 			Module:    "Module",
-			Blocked:   utils.IsBlockingEnabled(),
+			Blocked:   config.IsBlockingEnabled(),
```
bitterpanda63, line 83: we dont use serviceConfig here?

bitterpanda63: nvm we do

bitterpanda63 (approved):

PR #69, opened 2025-10-14
feat: use shared http client with timeout

tomaisthorpe:
Default `http.Client` doesn't have a timeout, should make sure one is always set.

### DIFF
--- a/internal/agent/cloud/event.go
+++ b/internal/agent/cloud/event.go
@@ -14,7 +14,7 @@ import (
 )
 
 var httpClient = &http.Client{
-	Timeout: 5 * time.Second,
+	Timeout: 30 * time.Second,
 }
 
 func SendCloudRequest(endpoint string, route string, method string, payload any) ([]byte, error) {
--- a/internal/agent/cloud/event.go
+++ b/internal/agent/cloud/event.go
@@ -7,12 +7,17 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"time"
 
 	"github.com/AikidoSec/firewall-go/internal/agent/config"
 	"github.com/AikidoSec/firewall-go/internal/agent/log"
 )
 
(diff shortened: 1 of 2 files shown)


internal/agent/cloud/event.go
```diff
@@ -24,43 +29,43 @@ func SendCloudRequest(endpoint string, route string, method string, payload inte
...
-		log.Infof("Sending %s request to %s with content %s", method, apiEndpoint, jsonData)
-
-		req, err = http.NewRequest(method, apiEndpoint, bytes.NewBuffer(jsonData))
-	} else {
-		log.Infof("Sending %s request to %s", method, apiEndpoint)
-		req, err = http.NewRequest(method, apiEndpoint, nil)
+		body = bytes.NewBuffer(jsonData)
 	}
 
+	log.Infof("Sending %s request to %s", method, apiEndpoint)
```
bitterpanda63: should be debug level no?

internal/agent/cloud/event.go
```diff
@@ -7,12 +7,17 @@ import (
...
 	"net/url"
+	"time"
 
 	"github.com/AikidoSec/firewall-go/internal/agent/config"
 	"github.com/AikidoSec/firewall-go/internal/agent/log"
 )
 
-func SendCloudRequest(endpoint string, route string, method string, payload interface{}) ([]byte, error) {
+var httpClient = &http.Client{
+	Timeout: 5 * time.Second,
```
hansott:
```suggestion
	Timeout: 30 * time.Second,
```

bitterpanda63 (changes requested):

internal/agent/cloud/event.go
tomaisthorpe: agreed, updated

bitterpanda63 (approved):

PR #88, opened 2025-10-20
Move log package to use slog

tomaisthorpe:
Currently doesn't support fields as part of logging, but will add that next. First step was using slog and allowing a logger to be set.

Example of this running with the current set of logs:
image

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -1,11 +1,12 @@
 package log
 
 import (
+	"context"
 	"errors"
 	"fmt"
-	"log"
+	"log/slog"
 	"os"
-	"time"
+	"strings"
 )
 
 type LogLevel int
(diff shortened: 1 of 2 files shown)


internal/log/log.go
```diff
@@ -18,50 +19,38 @@ const (
...
 }
 
 func logMessage(level LogLevel, args ...any) {
-	if level >= currentLogLevel {
-		formatter := &AikidoFormatter{}
-		message := fmt.Sprint(args...)
-		formattedMessage := formatter.Format(level, message)
-		Logger.Print(formattedMessage)
-	}
+	message := fmt.Sprint(args...)
```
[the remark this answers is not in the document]
tomaisthorpe: In theory you can call [`Enabled`](https://pkg.go.dev/log/slog#Logger.Enabled) to check if the log should be logged, but the internal methods of `slog` do that anyway, so not really necessary:
https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/log/slog/logger.go;l=244

internal/log/log.go
```diff
@@ -18,50 +19,38 @@ const (
 )
 
 var (
-	currentLogLevel = ErrorLevel
-	Logger          = log.New(os.Stdout, "", 0)
-	cliLogging      = true
+	levelVar  slog.LevelVar
+	logger    *slog.Logger
+	logWriter = os.Stdout
+	logFormat = "text" // "text" or "json"
```
bitterpanda63, line 25: feels like leftover comment, actually weird that codequal didnt pick up, can you add as thread?

tomaisthorpe: it's more of an internal comment just to show what the string can hold. It's validated from user input, so can only ever be those two values.

internal/log/log.go
```diff
@@ -97,17 +86,58 @@ func Errorf(format string, args ...any) {
...
+	case "json":
+		logFormat = "json"
+	default:
+		return errors.New("invalid log format")
+	}
+	rebuildLogger()
+	return nil
+}
+
+func toSlogLevel(level LogLevel) slog.Level {
```
bitterpanda63, line 130: could we not just remove our own "LogLevel" implementation?

tomaisthorpe: I think in theory we could remove it, but keeping the custom one has a few benefits:
- consumers don't _need_ to know we use slog internally
- more flexibility later if we want different behaviour, we already have the thing with `AIKIDO_DEBUG` which sets this to debug overriding everything else, maybe want to do something here?
- bit of a cleaner API

bitterpanda63: mm alright lets keep indeed the separation

bitterpanda63 (approved):

PR #94, opened 2025-10-22
feat: use structured logs

tomaisthorpe:
- Move all logs to use fields instead of string formatting
- Remove unused format methods

I haven't reviewed the log messages yet, just converting to structure logs.

### DIFF
--- a/internal/agent/config/service_config.go
+++ b/internal/agent/config/service_config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"log/slog"
 	"regexp"
 	"sync"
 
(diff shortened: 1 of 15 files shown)


internal/agent/config/service_config.go
```diff
@@ -67,8 +70,9 @@ func buildIPBlocklist(name, description string, ipsList []string) IPBlockList {
 		}
 	}
 
-	log.Debugf("%s (v4): %v", name, ipBlocklist.TrieV4)
-	log.Debugf("%s (v6): %v", name, ipBlocklist.TrieV6)
+	log.Debug("IP blocklist (v4)", slog.String("name", name), slog.Any("trie", ipBlocklist.TrieV4))
```
bitterpanda63: let's rm? these logs are not in other languages and are very spammy imo (even for debug logs)

bitterpanda63 (approved):

PR #97, opened 2025-10-24
Switch `zen-internals` to use WASM

tomaisthorpe:
- Added benchmark for detect sql injection
- Switched to WASM for zen internals
- Adjusted make file to fetch WASM binary and checksum
- Added runtime checksum verification
- Added test checksum verification for CI

Note: currently using `sync.Pool`, we probably need a pre-warmed pool, but will investigate this once we have it running in a long running application.


internal/main.go
```diff
@@ -1,8 +1,22 @@
...
 
-import "github.com/AikidoSec/firewall-go/internal/vulnerabilities/zeninternals"
+import (
+	"log/slog"
+
+	"github.com/AikidoSec/firewall-go/internal/log"
+	"github.com/AikidoSec/firewall-go/internal/vulnerabilities/zeninternals"
+)
+
+// Init initializes the firewall system by loading the zen-internals library
```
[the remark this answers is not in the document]
tomaisthorpe: Don't really agree with this, as an exported function, I would want to just know why an error is returned which this does.

internal/vulnerabilities/zeninternals/zen_internals.go
```diff
@@ -1,100 +1,215 @@
...
 func DetectSQLInjection(query string, userInput string, dialect int) int {
-	if detectSQLInjection == nil {
+	ctx := context.Background()
+
+	inst, ok := wasmPool.Get().(*wasmInstance)
+	if !ok || inst == nil {
+		log.Error("Failed to get WASM instance from pool")
 		return 0
 	}
+	defer wasmPool.Put(inst)
```
bitterpanda63: what does this do exactly? might need some comments since this is a bit complex

tomaisthorpe: It just puts the instance back in the pool, I think a comment would just be restating what the code is saying

internal/vulnerabilities/zeninternals/zen_internals.go
```diff
@@ -1,100 +1,215 @@
...
+		return 0, fmt.Errorf("detect_sql_injection returned value too large: %d", result)
+	}
+	return int32(result), nil
+}
+
+// allocateAndWriteString allocates memory for a string, writes it to WASM memory, and returns the pointer and length
+// The caller is responsible for freeing the memory using the returned free function
+func allocateAndWriteString(ctx context.Context, memory api.Memory, alloc, free api.Function, data []byte, name string) (uint32, uint64, func(), error) {
+	// Check for potential overflow in string length
+	const maxStringLen = 1 << 30 // 1GB limit to prevent overflow
```
bitterpanda63, line 179: Would block the query at this point, 1GB is pretty insane right? Let's make sure we don't create a bypass because of this

internal/vulnerabilities/zeninternals/zen_internals.go
```diff
@@ -1,100 +1,215 @@
...
+		return 0, 0, nil, fmt.Errorf("%s pointer too large: %d", name, ptrRaw)
+	}
+	ptr := uint32(ptrRaw)
+
+	// Write data to WASM memory
+	if !memory.Write(ptr, data) {
+		return 0, 0, nil, fmt.Errorf("failed to write %s to memory", name)
+	}
+
+	// Create cleanup function
```
bitterpanda63: Let's make sure we say why we are cleaninng up -> i.e. we created pointers ourselves, so wasm gc will not workf or that

internal/vulnerabilities/zeninternals/zen_internals.go
```diff
@@ -1,100 +1,215 @@
...
+	cleanup := func() {
+		if _, freeErr := free.Call(ctx, uint64(ptr), dataLen); freeErr != nil {
+			// Log error but don't fail the main operation
+			log.Error("Failed to free memory", slog.String("name", name), slog.Any("error", freeErr))
```
bitterpanda63: do we have any safeguards against oom crashes?

tomaisthorpe: In it's current implementation, if we can't free the memory, it would eventually get GC'd as this is only within the WASM instance, that just depends on the pool. One option is just not return the instance to the pool if we can't free memory, that would cause it to get GC'd much faster and reduce chance of OOM issues.

internal/vulnerabilities/zeninternals/zen_internals.go
bitterpanda63: ✅

internal/vulnerabilities/zeninternals/zen_internals.go
tomaisthorpe: Implemented the above!

bitterpanda63 (approved):

PR #101, opened 2025-10-27
feat: support monitoring mode

tomaisthorpe:
Updated 'scanSource' to not return an error if monitoring mode only is enabled.

### DIFF
--- a/internal/vulnerabilities/attack.go
+++ b/internal/vulnerabilities/attack.go
@@ -61,7 +61,6 @@ func getHeaders(context *request.Context) map[string][]string {
 	return headers
 }
 
-// getAttackDetected constructs the DetectedAttack struct to be to the Agent
 func getAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types.DetectedAttack {
 	reqCtx := request.GetContext(ctx)
 	if reqCtx == nil {
--- a/internal/vulnerabilities/attack.go
+++ b/internal/vulnerabilities/attack.go
@@ -107,11 +107,11 @@ func onInterceptorResult(ctx context.Context, res *InterceptorResult) error {
 		return nil
 	}
 
-	atk := getAttackDetected(ctx, *res)
-	go agent.OnAttackDetected(atk)
+	attack := getAttackDetected(ctx, *res)
+	go agent.OnAttackDetected(attack)
 
-	// If blocking is enabled, continue as normal after reporting the attack.
-	if atk == nil || !atk.Attack.Blocked {
+	// If blocking is disabled, continue as normal after reporting the attack.
+	if attack == nil || !attack.Attack.Blocked {
 		return nil
 	}
 
--- a/internal/vulnerabilities/attack.go
+++ b/internal/vulnerabilities/attack.go
@@ -23,7 +23,7 @@ const (
 	KindSSRF           AttackKind = "ssrf"
 )
 
-func GetDisplayNameForAttackKind(kind AttackKind) string {
+func getDisplayNameForAttackKind(kind AttackKind) string {
 	switch kind {
 	case KindSQLInjection:
 		return "an SQL injection"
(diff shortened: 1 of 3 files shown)


internal/vulnerabilities/attack.go
```diff
@@ -23,7 +23,7 @@ const (
 	KindSSRF           AttackKind = "ssrf"
 )
 
-func GetDisplayNameForAttackKind(kind AttackKind) string {
+func getDisplayNameForAttackKind(kind AttackKind) string {
```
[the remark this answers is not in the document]
tomaisthorpe: Purposefully unexported it.

internal/vulnerabilities/main.go
```diff
@@ -28,26 +29,30 @@ func Scan(ctx context.Context, operation string, vulnerability Vulnerability, ar
...
+
+	detectedAttack, err = scanSource(ctx, "body", reqCtx.Body, operation, vulnerability, args)
+	if detectedAttack || err != nil {
 		return err
 	}
 	return nil
 }
 
-func ScanSource(ctx context.Context, source string, sourceData any, operation string, vulnerability Vulnerability, args []string) error {
+// scanSource returns a boolean if an attack was detected and an error to be returned to the sink if the attack should be blocked
```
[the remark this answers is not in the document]
tomaisthorpe: I think this is obvious when you look at where this is used, and it does explain why an error is returned.

internal/vulnerabilities/attack.go
```diff
@@ -61,8 +61,7 @@ func getHeaders(context *request.Context) map[string][]string {
 	return headers
 }
 
-// GetAttackDetected constructs the DetectedAttack struct to be to the Agent
-func GetAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types.DetectedAttack {
+func getAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types.DetectedAttack {
```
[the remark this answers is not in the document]
tomaisthorpe: I don't think there's any particularly interesting behaviour to comment, it's just creating a struct, no real edge-cases.

internal/vulnerabilities/attack.go
```diff
@@ -93,36 +92,28 @@ func GetAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types
...
+	if res == nil {
+		return nil
 	}
-	return string(actionJSON)
-}
 
-func GetAttackDetectedAction(result InterceptorResult) string {
-	return GetThrowAction(BuildAttackDetectedMessage(result), -1)
-}
+	atk := getAttackDetected(ctx, *res)
```
bitterpanda63: atk is a bit of a weird abbreviation?

internal/vulnerabilities/attack.go
```diff
@@ -93,36 +92,28 @@ func GetAttackDetected(ctx context.Context, res InterceptorResult) *aikido_types
...
+	go agent.OnAttackDetected(atk)
 
-func ReportAttackDetected(ctx context.Context, res *InterceptorResult) string {
-	if res == nil {
-		return ""
+	// If blocking is enabled, continue as normal after reporting the attack.
```
bitterpanda63: mm not understanding this comment, guessing you mean if blocking is disabled we continue as normal?

internal/vulnerabilities/main.go
```diff
@@ -28,26 +29,30 @@ func Scan(ctx context.Context, operation string, vulnerability Vulnerability, ar
 		return nil
 	}
 
-	err := ScanSource(ctx, "query", reqCtx.Query, operation, vulnerability, args)
-	if err != nil {
+	detectedAttack, err := scanSource(ctx, "query", reqCtx.Query, operation, vulnerability, args)
+	if detectedAttack || err != nil {
```
bitterpanda63: mm are we sure this should be an or, I guess it sort of makes some sense to not keep running attack scanning, although I think it makes more sense to keep running it?
- non-blocking mode is used to make sure it will be safe to deploy zen in prod apps, so we want any possible false positives
- makes the code easier to understand if its simply err != nil

bitterpanda63: on the other hand in most other agents if we detect an attack from one source, we just stop the algorithm wheter its in blocking or monitoring mode

internal/vulnerabilities/attack.go
tomaisthorpe: 🤦

internal/vulnerabilities/main.go
tomaisthorpe: I'm not sure the best option for this. I was wondering if we should run all these `scanSource` in parallel, which would mean it's possible we would find multiple attacks before we collate the results.

Happy to just make it so we keep scanning in monitoring only mode? I did it like this to reduce extra work on a request if we had spotted an attack, but it's not much extra really.

bitterpanda63 (approved):

PR #103, opened 2025-10-27
feat: use cloud client config for endpoints and token

tomaisthorpe:
- Using client config for endpoint and tokens to make easier to setup mock servers
- Removed global constants to the cloud package where they are used
- Renamed configEndpoint > realtimeEndpoint for consistency


internal/agent/agent.go
```diff
@@ -31,7 +31,11 @@ func Init(environmentConfig *aikido_types.EnvironmentConfigData, aikidoConfig *a
 		return false
 	}
 
-	cloudClient = cloud.NewClient(&cloud.ClientConfig{})
+	cloudClient = cloud.NewClient(&cloud.ClientConfig{
+		Token:            aikidoConfig.Token,
+		APIEndpoint:      globals.EnvironmentConfig.Endpoint,
```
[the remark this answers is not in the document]
tomaisthorpe: There's no reasonable way around this currently. The environment config still sits within those globals.

internal/agent/cloud/event.go
```diff
@@ -10,15 +10,18 @@ import (
...
 	"net/url"
 
-	"github.com/AikidoSec/firewall-go/internal/agent/config"
 	"github.com/AikidoSec/firewall-go/internal/log"
 )
 
 var ErrNoTokenSet = errors.New("no token set")
 
+const (
+	eventsAPIMethod = "POST"
```
bitterpanda63, line 19: where are these used btw?

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,84 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+	} else if heartbeatIntervalInMS >= minHeartbeatIntervalInMS {
+		log.Info("Resetting HeartBeatTicker!", slog.Int("interval", heartbeatIntervalInMS))
+		heartBeatTicker.Reset(time.Duration(heartbeatIntervalInMS) * time.Millisecond)
+	}
+}
+
+func updateRateLimitingConfig(endpoints []aikido_types.Endpoint) {
+	endpointConfigs := make([]ratelimiting.EndpointConfig, len(endpoints))
+	for i, endpoint := range endpoints {
+		endpointConfigs[i] = ratelimiting.EndpointConfig{
```
bitterpanda63, line 111: is there really no easier way of doing the update?

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,84 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+	resetHeartbeatTicker(cloudConfig.HeartbeatIntervalInMS, cloudConfig.ReceivedAnyStats)
+	updateRateLimitingConfig(cloudConfig.Endpoints)
+
+	config.UpdateServiceConfig(cloudConfig)
+	return true
+}
+
+func resetHeartbeatTicker(heartbeatIntervalInMS int, receivedAnyStats bool) {
+	if !receivedAnyStats {
+		log.Info("Resetting HeartBeatTicker to 1m!")
```
bitterpanda63: def not info-level lets downgrade or remove even

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,84 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+	}
+	cloudConfig.BlockedUserAgents = tempListsConfig.BlockedUserAgents
+	return true
+}
+
+func (c *Client) storeCloudConfig(configResponse []byte) bool {
+	cloudConfig := &aikido_types.CloudConfigData{}
+	err := json.Unmarshal(configResponse, &cloudConfig)
+	if err != nil {
+		log.Warn("Failed to unmarshal cloud config!")
```
bitterpanda63: feels like you would maybe want to know the response so we can go into debugging mode fo why it failed to marshal? could add as 2nd debug log also

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,84 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+		return false
+	}
+	if cloudConfig.ConfigUpdatedAt <= config.GetCloudConfigUpdatedAt() {
+		log.Debug("ConfigUpdatedAt is the same!")
```
bitterpanda63: not useful, literally just a log saying "no action" every 1min
```suggestion
```

internal/agent/cloud/event.go
bitterpanda63: nevermind, saw the _ import just now

internal/agent/cloud/config.go
tomaisthorpe: Unfortunately not! If we're mapping from one type to another and wanting to clone it, this is the only real option

internal/agent/cloud/config.go
tomaisthorpe: I've added the error as a field for the log, it should be enough to debug anything here.

internal/agent/cloud/event.go
```diff
@@ -10,15 +10,18 @@ import (
 	"net/http"
 	"net/url"
 
-	"github.com/AikidoSec/firewall-go/internal/agent/config"
 	"github.com/AikidoSec/firewall-go/internal/log"
 )
 
 var ErrNoTokenSet = errors.New("no token set")
 
+const (
```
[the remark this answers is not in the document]
tomaisthorpe: These are used in the files that call the events API. I think the naming is probably fine.

internal/agent/aikido_types/init_data.go
```diff
@@ -3,12 +3,12 @@ package aikido_types
...
-	PlatformName    string `json:"platform_name"`             // Platform name (fpm-fcgi, cli-server, ...)
-	PlatformVersion string `json:"platform_version"`          // Language version
-	Endpoint        string `json:"endpoint,omitempty"`        // default: 'https://guard.aikido.dev/'
-	ConfigEndpoint  string `json:"config_endpoint,omitempty"` // default: 'https://runtime.aikido.dev/'
-	Library         string `json:"library,omitempty"`         // default: 'firewall-php'
-	Version         string `json:"version,omitempty"`         // Version of the agent
+	PlatformName     string `json:"platform_name"`               // Platform name (fpm-fcgi, cli-server, ...)
+	PlatformVersion  string `json:"platform_version"`            // Language version
+	Endpoint         string `json:"endpoint,omitempty"`          // default: 'https://guard.aikido.dev/'
+	RealtimeEndpoint string `json:"realtime_endpoint,omitempty"` // default: 'https://runtime.aikido.dev/'
```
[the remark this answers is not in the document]
tomaisthorpe: No issue here, it's actually not converted to JSON at any point. Will remove those struct tags later.

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,88 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+	}
+	cloudConfig.BlockedUserAgents = tempListsConfig.BlockedUserAgents
+	return true
+}
+
+// storeCloudConfig applies cloud configuration if newer than the current
+// version. Returns false only if unmarshaling fails.
+func (c *Client) storeCloudConfig(configResponse []byte) bool {
+	cloudConfig := &aikido_types.CloudConfigData{}
+	err := json.Unmarshal(configResponse, &cloudConfig)
```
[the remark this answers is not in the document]
tomaisthorpe: Reported as false positive.

internal/agent/cloud/config.go
```diff
@@ -25,11 +39,88 @@ func (c *Client) CheckConfigUpdatedAt() {
...
+	config.UpdateServiceConfig(cloudConfig)
+	return true
+}
+
+// resetHeartbeatTicker sets heartbeat interval to 1 minute initially,
+// then uses the provided interval if it meets the minimum threshold.
+func resetHeartbeatTicker(heartbeatIntervalInMS int, receivedAnyStats bool) {
+	if !receivedAnyStats {
+		heartBeatTicker.Reset(1 * time.Minute)
+	} else if heartbeatIntervalInMS >= minHeartbeatIntervalInMS {
```
bitterpanda63, line 100: wouldnt we want to default here to the 10min if the heartbeat given by core is invalid? or do we just remain on the previous heartbeat time?

tomaisthorpe: I think it's probably fine as it is, it just means we keep the previous heartbeat is we get an invalid value.

bitterpanda63 (approved):

PR #116, opened 2025-10-29
Add zen-go cmd for setting up project

tomaisthorpe:
- Add zen-go cmd for quick setup of new projects
  - Added under separate package for speedy install
- Add codecov flags and uploaded main, zen-go and instrumentation coverage
- Set version to 0.0.0 to match the fact we have no tags yet.


cmd/zen-go/main.go
```diff
@@ -0,0 +1,107 @@
...
+	case "help", "-h", "--help":
+		printUsage()
+	default:
+		fmt.Fprintf(os.Stderr, "Unknown command: %s\n\n", os.Args[1])
+		printUsage()
+		os.Exit(1)
+	}
+}
+
+func initCommand() error {
```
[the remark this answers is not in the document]
tomaisthorpe: I don't think this is necessary, the behaviour is documented by the app itself.

Makefile
```diff
@@ -14,13 +14,24 @@ install-tools:
...
+	@echo "✅ Main module tests completed successfully"
 	@echo "Coverage report saved to coverage.out"
 	@go tool cover -func=coverage.out | grep total | awk '{print "Total coverage: " $$3}'
 
+.PHONY: test-zen-go
+test-zen-go:
+	@echo "Running zen-go CLI tests with gotestsum"
+	@cd cmd/zen-go && gotestsum --format pkgname -- -race -coverprofile=coverage.out -covermode=atomic ./...
+	@echo "✅ zen-go tests completed successfully"
+	@cd cmd/zen-go && go tool cover -func=coverage.out | grep total | awk '{print "zen-go coverage: " $$3}'
```
bitterpanda63:
```suggestion
	@cd cmd/zen-go && go tool cover -func=coverage.out | grep total | awk '{print "cmd/zen-go coverage: " $$3}'
```
might make more sense, zen-go could also just be name of main repo, up to you

cmd/zen-go/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const version = "0.0.0"
+
+const toolsFileTemplate = `// This file was created by 'zen-go init', and is used to ensure the
```
bitterpanda63: can we move this & the initCommand logic to separate file?

cmd/zen-go/main.go
```diff
@@ -0,0 +1,107 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+const version = "0.0.0"
```
bitterpanda63, line 8: would we also update this one during release?

bitterpanda63: or I guess, when merging into main 🤔

tomaisthorpe: When we do a release, we would just bump these versions and then create a tag on the commit where we bumped the versions.

cmd/zen-go/main.go
```diff
@@ -0,0 +1,54 @@
...
+		printUsage()
+	default:
+		fmt.Fprintf(os.Stderr, "Unknown command: %s\n\n", os.Args[1])
+		printUsage()
+		os.Exit(1)
+	}
+}
+
+func printUsage() {
+	fmt.Println("zen-go - Aikido Zen CLI tool for Go")
```
bitterpanda63, line 34: could we extract this zen-go into a constant, just making sure it will be easy to change if we want to? (i.e. if we provide alias in future bc of rename)

cmd/zen-go/cmd_init.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	// Ensures Aikido Zen instrumentation is present in go.mod
+	// Do not remove this unless you want to stop using Aikido.
+	_ "github.com/AikidoSec/firewall-go/instrumentation"
+	_ "github.com/DataDog/orchestrion" // integration
+)
+`
+
+func initCommand() error {
+	// Parse flags
```
bitterpanda63:
```suggestion
	// Parse force flag: if enabled will overwrite existing files if necessary
```

cmd/zen-go/cmd_init.go
```diff
@@ -0,0 +1,59 @@
...
+	if err := os.WriteFile(filename, []byte(toolsFileTemplate), 0o644); err != nil {
+		return fmt.Errorf("failed to create %s: %w", filename, err)
+	}
+
+	absPath, _ := filepath.Abs(filename)
+	fmt.Printf("✓ Created %s\n", filename)
+	fmt.Printf("  %s\n\n", absPath)
+	fmt.Println("Next steps:")
+	fmt.Println("  1. Run 'go mod tidy' to update your dependencies")
+	fmt.Println("  2. Build with 'orchestrion go build' to enable instrumentation")
```
bitterpanda63, line 56: should we also make this smth like `zen-go go build`?

tomaisthorpe: I think it could be something we look at in future, it would at least make the library agnostic to the compile time instrumentation we use, for now though, I think it's too much effort.

cmd/zen-go/main.go
tomaisthorpe: I'd prefer not to extract to a constant since it would require string formatting throughout. I'm also considering switching to a CLI package (like cobra/urfave/cli) which would handle this automatically with a single config point. Originally avoided it due to dependency concerns, but the separate go.mod makes that less of an issue.

cmd/zen-go/cmd_init.go
bitterpanda63: can be separate PR

cmd/zen-go/main.go
bitterpanda63: makes sense, alright will resolve for now

bitterpanda63 (approved):

PR #137, opened 2025-11-04
Add os/exec sink with shell injection detection

tomaisthorpe:
- Add new sink for os/exec for detecting shell injection.
  - The detection and block happens when the command is actually run through `Run`/`Start`/etc.
- Add helpers to check if it's a shell invocation, this is the only way to trigger shell injection.


internal/vulnerabilities/shellinjection/parse_shell_command.go
```diff
@@ -0,0 +1,37 @@
+package shellinjection
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// IsShellCommand checks if the given program path is a shell interpreter.
```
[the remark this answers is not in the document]
tomaisthorpe: It's nothing about specific shells, it's checking if something is a shell command, I think this is fine to me.

internal/vulnerabilities/shellinjection/parse_shell_command.go
```diff
@@ -0,0 +1,37 @@
...
+// It returns true for common shells like sh, bash, zsh, etc.
+func IsShellCommand(program string) bool {
+	shell := strings.ToLower(filepath.Base(program))
+
+	switch shell {
```
bitterpanda63, line 13: seems a bit dirty to do this with a switch case, would do more readable if and a constant of shells so that we could expand if necessary?

internal/vulnerabilities/shellinjection/parse_shell_command.go
```diff
@@ -0,0 +1,37 @@
...
+// ExtractShellCommandString extracts the command string from shell invocation arguments.
+// It looks for the -c flag and returns everything after it as the command string
+// to be interpreted by the shell.
+func ExtractShellCommandString(args []string) []string {
+	for i := 1; i < len(args); i++ {
+		flag := args[i]
+		if flag == "-c" {
+			// Return everything after the flag
+			if i+1 < len(args) {
+				return args[i+1:]
```
bitterpanda63, line 30: wouldnt it just be the argument after the -c? Or does it still work even if its split up into multiple arguments?

internal/vulnerabilities/shellinjection/parse_shell_command.go
HatiCode: For simple validation the switch statement is just fine, clean/idiomatic and actually performs better than if statement. Readability wise I guess both options are legit

instrumentation/sinks/os/exec/examine.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(args) == 0 {
+		return nil
+	}
+
+	// Only shell invocations can have injection vulnerabilities
+	if !shellinjection.IsShellCommand(args[0]) {
+		return nil
+	}
+
+	// Extract the command string that will be interpreted by the shell
```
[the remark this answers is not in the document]
tomaisthorpe: I think this comment is useful. The command above it states that only shell injection is possible with a shell command, and then the next step is extracting the commands that the shell will interpret.

internal/vulnerabilities/shellinjection/parse_shell_command.go
tomaisthorpe: Are you meaning if they run something like:
```go
exec.Command("sh", "-c", "ls", userInput)
```
In that case, I don't think it's possible to do shell injection.

However, I think we want to be able to detect cases with positional arguments, like:
```go
exec.Command("sh", "-c", "ls $0", userInput)
```

bitterpanda63: userInput -> ".; rm -rf /"
basically sh is not fool-proof, you can quickly escalate with args

bitterpanda63: `sh -c "ls $1" ".; touch flag.txt;"`
not executing code

bitterpanda63: `sh -c "ls -la" -c "touch flag.txt"`
only seems to execute the first command

tomaisthorpe: Currently, only case we can think of where this is an issue is:
```go
cmd := exec.Command("sh", "-c", "$0", userInput)
```

Unlikely, but worth covering either way.

hansott (approved):

PR #149, opened 2025-11-07
Add net/http source for ServeMux

tomaisthorpe:
Added first version of the source for `net/http.ServeMux`, should support simple use-cases.

For the multi-part form parsing, I used 32MB as a max size, which is just copied from `gin` and `echo`:
https://github.com/gin-gonic/gin/blob/c3d5a28ed6d3849da820195b6774d212bcc038a9/gin.go#L26

There's a known issue with nested ServeMux, but didn't look at fixing that just yet to keep the PR a little simpler.


instrumentation/sources/net/http/body.go
```diff
@@ -0,0 +1,37 @@
...
+	if bodyFromForm != nil {
+		return bodyFromForm
+	}
+
+	// No use-able data found, returning nil :
+	return nil
+}
+
+func tryExtractFormBody(r *http.Request) url.Values {
+	if err := r.ParseMultipartForm(32 << 20); err != nil {
```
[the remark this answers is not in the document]
tomaisthorpe: this is expected, if it's not a multipart form, then there's nothing for us to do.

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		// Wrap the ResponseWriter to capture the status code
+		recorder := &statusRecorder{
+			ResponseWriter: w,
+		}
+
+		orig(recorder, wrappedR)
+
+		zenhttp.OnPostRequest(ctx, recorder.statusCode)
```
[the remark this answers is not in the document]
tomaisthorpe: That's handled within OnPostRequest.

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+	"sync/atomic"
+
+	zenhttp "github.com/AikidoSec/firewall-go/internal/http"
+	"github.com/AikidoSec/firewall-go/internal/request"
+)
+
+// Middleware sets the request contexts of incoming requests.
+// It is part of the automatic instrumentation and will be run before any other middleware.
+func Middleware(orig func(w http.ResponseWriter, r *http.Request)) func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
tomaisthorpe: I don't think this is valid, the point of the function is to return handler func, it's pretty standard behaviour.

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,71 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		ip := r.RemoteAddr
+
+		pattern := r.Pattern
+		// ServeMux patterns can start with the method and/or host
+		// We only care about the path, we need to strip it
+		pattern = strings.TrimPrefix(pattern, r.Method+" ")
+
+		// Strip host prefix (e.g., "example.com/path")
```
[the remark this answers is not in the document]
tomaisthorpe: It mentioned why we care about this 2 lines above.

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/AikidoSec/firewall-go/internal/request"
+)
+
+// Middleware sets the request contexts of incoming requests.
+// It is part of the automatic instrumentation and will be run before any other middleware.
+func Middleware(orig func(w http.ResponseWriter, r *http.Request)) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ip := r.RemoteAddr
+
+		pattern := r.Pattern
```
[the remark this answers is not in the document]
tomaisthorpe: False positive

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,125 @@
...
+		// ServeMux patterns can start with the method and/or host
```
bitterpanda63, line 21: "can", but we always strip the first part? (strings.Index(pattern, "/"))

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,125 @@
...
+		// Strip host prefix (e.g., "example.com/path")
+		if idx := strings.Index(pattern, "/"); idx > 0 {
+			pattern = pattern[idx:]
+		}
+
+		ctx := request.SetContext(r.Context(), r, request.ContextData{
+			Source:        "http.ServeMux",
+			Route:         pattern,
+			RouteParams:   extractRouteParams(r, pattern),
+			RemoteAddress: &ip,
```
bitterpanda63, line 36: does servemux resolve x-forwarded-for? otherwise would add x-forwarded-for logic in next pr

bitterpanda63: or is this handled in the SetContext itself

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,125 @@
...
+
+		res := zenhttp.OnInitRequest(ctx)
+		if res != nil {
+			w.WriteHeader(res.StatusCode)
+			_, _ = w.Write([]byte(res.Message))
+			return
+		}
+
+		// Wrap the ResponseWriter to capture the status code
+		recorder := &statusRecorder{
```
bitterpanda63, line 50: is it not safer to add a 2nd entrypoint with orchestrion?

instrumentation/sources/net/http/middleware.go
tomaisthorpe: It does not. We do have to be careful with that, as it could potentially allow bypasses for allowed IPs, etc.

Gin gets around this by configuring 'trusted' proxies. I'll add it to my todo list to look how to do this safely.

instrumentation/sources/net/http/middleware.go
tomaisthorpe: We're not always stripping it, it will only strip it is there's a host there. If there's no host, then `idx == 0`, so nothing changes.

instrumentation/sources/net/http/middleware.go
tomaisthorpe: What safety issues are you thinking about?

And I think adding another orchestrion instrumentation for this would complicate it a fair bit, I think the interface is a simple way to deal with the problem of capturing the status code.

instrumentation/sources/net/http/middleware.go
bitterpanda63: mm are the slashes in the path urlencoded then? what happens if path is e.g. test/route/2

instrumentation/sources/net/http/middleware.go
bitterpanda63: mm alright, we have some code in the other agents to deal with that btw, best if its similar

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+type statusRecorder struct {
+	writer     http.ResponseWriter
+	statusCode int
+	written    atomic.Bool
+}
+
+func (r *statusRecorder) Unwrap() http.ResponseWriter {
+	return r.writer
```
bitterpanda63, line 74: add comment when this is used, lets make sure it never happens for http apps right
api specs are important

instrumentation/sources/net/http/middleware.go
```diff
@@ -0,0 +1,134 @@
...
+func (r *statusRecorder) Header() http.Header {
+	return r.writer.Header()
+}
+
+type requestParser struct {
+	req *http.Request
+}
+
+func (rp *requestParser) MultipartForm() (*multipart.Form, error) {
+	err := rp.req.ParseMultipartForm(32 << 20)
```
bitterpanda63, line 105: can we add src link (permalink)

tomaisthorpe: added

instrumentation/sources/net/http/middleware.go
tomaisthorpe: added debug log

bitterpanda63 (approved):

PR #155, opened 2025-11-10
Return error when attack is blocked on DB.ExecContext

tomaisthorpe:
Moving the SQL sink to return errors instead of panics.
This just returns an error which can be handled more gracefully within the service instead of panics.
The error is wrapped in `ErrAttackBlocked` incase the consumer wishes to handle them in a specific way.

### DIFF
--- a/zen/main.go
+++ b/zen/main.go
@@ -3,6 +3,7 @@
 package zen
 
 import (
+	"errors"
 	"log/slog"
 	"os"
 	"runtime"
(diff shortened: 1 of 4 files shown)


zen/main.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/AikidoSec/firewall-go/internal/log"
 )
 
+var ErrAttackBlocked = errors.New("zen blocked attack")
```
bitterpanda63: Error should be per attack type

bitterpanda63: https://github.com/AikidoSec/firewall-java/blob/1d87e9e314b0587d1e06452722f1155405f72947/agent_api/src/main/java/dev/aikido/agent_api/vulnerabilities/ssrf/SSRFException.java#L6
e.g.

bitterpanda63 (approved):

PR #161, opened 2025-11-12
Add tests for rate limiting and make it more testable

tomaisthorpe:
Continuing to reduce overall reliance on globals for state. Once config has been refactored, I think we should be able to get rid of the global rate limiter instance too and just keep an instance of it in the agent.

### DIFF
--- a/internal/agent/ratelimiting/rate_limiting.go
+++ b/internal/agent/ratelimiting/rate_limiting.go
@@ -44,17 +44,24 @@ type Status struct {
 	Trigger string
 }
 
-var (
+type RateLimiter struct {
 	// rateLimitingMap holds the current rate limiting state for each configured route
-	rateLimitingMap = make(map[Key]*Value)
+	rateLimitingMap map[Key]*Value
 
-	// mutex is used to sync access across the go routines
-	mutex sync.RWMutex
+	mu sync.RWMutex
 
 	// Channel and Ticker for the rate limiting background routine
-	channel = make(chan struct{})
-	ticker  = time.NewTicker(MinRateLimitingIntervalInMs * time.Millisecond)
-)
+	channel chan struct{}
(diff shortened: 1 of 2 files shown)


internal/agent/ratelimiting/rate_limiting.go
```diff
@@ -14,28 +14,28 @@ const (
...
 
-type Config struct {
+type rateLimitConfig struct {
 	MaxRequests         int
 	WindowSizeInMinutes int
 }
 
-// Counts tracks rate limiting counts for a specific entity (user or IP)
-type Counts struct {
+// entityCounts tracks rate limiting counts for a specific entity (user or IP)
```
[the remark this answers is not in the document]
tomaisthorpe: I think this is a useful comment for this struct, some of the rate limiting is a bit tough to follow.

internal/agent/ratelimiting/rate_limiting.go
```diff
@@ -44,19 +44,26 @@ type Status struct {
...
+	channel chan struct{}
+	ticker  *time.Ticker
+}
 
-func advanceQueuesForMap(config *Config, countsMap map[string]*Counts) {
+func New() *RateLimiter {
+	return &RateLimiter{
+		rateLimitingMap: make(map[endpointKey]*endpointData),
+		channel:         make(chan struct{}),
+		ticker:          time.NewTicker(MinRateLimitingIntervalInMs * time.Millisecond),
```
bitterpanda63, line 62: could we rm this logic?

tomaisthorpe: Simplifying this logic in:
https://github.com/AikidoSec/firewall-go/pull/163

Keeping this PR about testing for now.

bitterpanda63 (approved):

PR #168, opened 2025-11-14
Add state collector for hostnames, routes and middleware installed

tomaisthorpe:
Moves state into a separate package for easier testing and keeps responsibility of agent package smaller.

Stats data will be moved here, but there's a lot of cod, so splitting up the PRs. 

And will add users here after:
- https://github.com/AikidoSec/firewall-go/pull/169


internal/agent/state/collector.go
```diff
@@ -0,0 +1,37 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/AikidoSec/firewall-go/internal/agent/aikido_types"
+)
+
+type Collector struct {
+	mu sync.Mutex
+
+	// List of outgoing hostnames, their ports and number of hits, collected from the requests
```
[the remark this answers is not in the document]
tomaisthorpe: I'd rather keep this here as the data structure is not obvious.

internal/agent/state/hostnames_test.go
```diff
@@ -25,17 +25,6 @@ func TestStoreDomain(t *testing.T) {
 			shouldStore:   true,
 			calls:         1,
 		},
-		{
```
bitterpanda63, line 28: can we re-add this test case?

tomaisthorpe: It was removed because there's another test case that does exactly the same thing:
> multiple calls to same domain and port

bitterpanda63 (approved):

PR #185, opened 2025-11-21
Add rate limiting by groups

tomaisthorpe:


zen/set_rate_limit_group.go
```diff
@@ -0,0 +1,27 @@
...
+
+// SetRateLimitGroup associates a group with the current request context which is used for rate limiting.
+// This function must be called before the Zen middleware is executed.
+func SetRateLimitGroup(ctx context.Context, id string) context.Context {
+	if len(id) == 0 {
+		log.Info("Group ID cannot be empty.")
+		return ctx
+	}
+
+	reqCtx := request.GetContext(ctx)
```
hansott, line 18: https://github.com/AikidoSec/firewall-node/blob/main/library/ratelimiting/group.ts#L72-L83
Might be interesting for the user to know? When the context isn't available? But the context might also be absent due to bypass IPs, right? (that's different in node)

tomaisthorpe: Yes! This log line doesn't make sense with the bypassed IPs. I've not actually implemented removing the context for bypassed IPs yet, but it would make this log line appear. It would probably make sense to return an error from this function for things like validating the group name, but not sure how we'd determine if it's missing the context due to bypassed IP.

hansott (approved):

zen/set_rate_limit_group.go
```diff
@@ -0,0 +1,27 @@
...
+
+	"github.com/AikidoSec/firewall-go/internal/log"
+	"github.com/AikidoSec/firewall-go/internal/request"
+)
+
+// SetRateLimitGroup associates a group with the current request context which is used for rate limiting.
+// This function must be called before the Zen middleware is executed.
+func SetRateLimitGroup(ctx context.Context, id string) context.Context {
+	if len(id) == 0 {
+		log.Info("Group ID cannot be empty.")
```
bitterpanda63, line 14: would make more sense to return as error here no?

zen/set_rate_limit_group.go
```diff
@@ -0,0 +1,27 @@
...
+		return ctx
+	}
+
+	reqCtx := request.GetContext(ctx)
+	if reqCtx == nil || reqCtx.HasMiddlewareExecuted() {
+		log.Info("zen.SetRateLimitGroup(...) must be called before the Zen middleware is executed.")
```
bitterpanda63, line 20: they could technically also be running the middleware multiple times

bitterpanda63 (approved):

zen/set_rate_limit_group.go
tomaisthorpe: I think for that it might be better that we log something if they run ShouldBlockRequest more than once, then we're not having to handle that case in both `SetUser` and `SetRateLimitGroup`?

zen/set_rate_limit_group.go
tomaisthorpe: Will fix in another PR along with `SetUser`.

PR #189, opened 2025-11-25
Add attack wave detection

tomaisthorpe:
Detection logic for the attack waves, code to actually trigger and report attack waves will come in the next PR.

Using an LRU cache to give upper limit to memory use

## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 13 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added attack wave detection with reporting and request integration

**⚡ Enhancements**
* Implemented simple LRU cache to bound memory and expiry
* Added suspicious filename/extension/directory lists and detection helpers

<sup>[More info](https://app.aikido.dev/featurebranch/scan/74062562?groupId=6)</sup>


internal/vulnerabilities/attackwave/lru_cache.go
```diff
@@ -0,0 +1,121 @@
...
+func newLRUCache(capacity int, ttl time.Duration) *lruCache {
+	return &lruCache{
+		capacity: capacity,
+		ttl:      ttl,
+		items:    make(map[string]*list.Element),
+		list:     list.New(),
+	}
+}
+
+func (c *lruCache) Get(key string) ([]time.Time, bool) {
```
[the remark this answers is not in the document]
tomaisthorpe: It's an LRU cache, not sure this is necessary!

internal/vulnerabilities/attackwave/patterns_directories.go
```diff
@@ -0,0 +1,62 @@
+package attackwave
+
+import "strings"
+
+// suspiciousDirectories contains directory names commonly targeted by web scanners
+var suspiciousDirectories map[string]bool
+
+func init() {
```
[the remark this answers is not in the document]
tomaisthorpe: This is fair, but want to keep these files separate to make them easier to check.

internal/vulnerabilities/attackwave/patterns_extensions.go
```diff
@@ -0,0 +1,27 @@
+package attackwave
+
+import "strings"
+
+// suspiciousExtensions contains file extensions commonly targeted by web scanners
+var suspiciousExtensions map[string]bool
+
+func init() {
```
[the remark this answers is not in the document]
tomaisthorpe: Answered above.

internal/vulnerabilities/attackwave/patterns_files.go
```diff
@@ -0,0 +1,317 @@
+package attackwave
+
+import "strings"
+
+// suspiciousFilenames contains file names commonly targeted by web scanners
+var suspiciousFilenames map[string]bool
+
+func init() {
```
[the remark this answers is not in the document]
tomaisthorpe: Answered above.

internal/vulnerabilities/attackwave/detector.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"time"
+
+	"github.com/AikidoSec/firewall-go/internal/request"
+	"github.com/AikidoSec/firewall-go/internal/slidingwindow"
+)
+
+// Detector tracks suspicious requests per IP and reports attack waves
+type Detector struct {
+	// attackWaveThreshold is the number of suspicious requests needed within the time frame to trigger an alert
```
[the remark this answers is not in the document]
tomaisthorpe: I'd rather the purpose of these fields be very clear.

internal/agent/agent_test.go
```diff
@@ -76,3 +79,6 @@ func TestCloudClient(t *testing.T) {
 		wg.Wait()
 	})
 }
+
+func TestOnAttackWaveDetected(t *testing.T) {
```
bitterpanda63: empty test case?

internal/vulnerabilities/attackwave/suspicious_query.go
```diff
@@ -0,0 +1,59 @@
...
+	// Collect all value slices from the map
+	valueSlices := slices.Collect(maps.Values(ctx.Query))
+
+	// Flatten and check
+	allValues := slices.Concat(valueSlices...)
+	return slices.ContainsFunc(allValues, containsDangerousPattern)
+}
+
+func containsDangerousPattern(str string) bool {
+	// Performance optimization: skip very short or very long strings
```
bitterpanda63, line 45: add importnat context : 
    // Some keywords like ../ are shorter than this min length check
    // However, they are part of a larger string in the most cases
    // e.g. ../etc/passwd or MD5(something)

internal/vulnerabilities/attackwave/suspicious_paths.go
```diff
@@ -0,0 +1,42 @@
...
+	segments := strings.Split(normalized, "/")
+
+	// Get the last segment (filename)
+	var filename string
+	if len(segments) > 0 {
+		filename = segments[len(segments)-1]
+	}
+
+	// Check if the filename is in our suspicious list
+	if filename != "" && suspiciousFilenames[filename] {
```
bitterpanda63, line 19: if the filename == "", then the path is just "" or "/" ?, maybe useful early return case to make it easier to read

internal/request/context.go
```diff
@@ -13,6 +13,7 @@ type DeferredBlock struct {
 
 type Context struct {
 	URL                string
+	Path               string
```
bitterpanda63, line 16: normally we just extract from URL no?

bitterpanda63: ok but is fine, is r.URL.Path

internal/request/http_context.go
```diff
@@ -44,6 +44,7 @@ func SetContext(ctx context.Context, r *http.Request, data ContextData) context.
 
 	c := &Context{
 		URL:                fullURL(r),
+		Path:               r.URL.Path,
```
bitterpanda63: can you add comment explaining this is w/o query params inside the path

bitterpanda63: fragments are also not in here right?

internal/http/on_post_request.go
```diff
@@ -32,6 +32,10 @@ func OnPostRequest(ctx context.Context, statusCode int) {
 		return
 	}
 
+	if agent.CheckAttackWave(reqCtx) {
```
bitterpanda63, line 21: maybe add a small comment specifying we run it here so that we can also report user

internal/agent/agent.go
```diff
@@ -134,6 +138,37 @@ func OnMiddlewareInstalled() {
...
+	UserAgent string
+	Source    string
+}
+
+func OnAttackWaveDetected(ctx *request.Context) {
+	if ctx == nil {
+		return
+	}
+
+	log.Debug("Reporting attack wave")
```
bitterpanda63, line 160: would be cleaner if its inside the Send itself no?

internal/vulnerabilities/attackwave/suspicious_paths.go
tomaisthorpe: It can't be an early return because we later check all the segments of the path. It's written in a slightly different way with a nested if on the node version:
https://github.com/AikidoSec/firewall-node/blob/33d4ddcc44bc01415e41d3eaf61c8b5a259a6a38/library/vulnerabilities/attack-wave-detection/isWebScanPath.ts#L29

internal/agent/agent.go
tomaisthorpe: This is following what the other methods do within `agent.go`, I think this makes more sense because some of these methods don't send anything. For example, the methods which are just storing stats.

internal/request/http_context.go
tomaisthorpe: You wouldn't see the fragments as this is on the server. You only get fragments in the url when you're sending a request as a client.

I've added comment for the path though.

internal/agent/agent_test.go
tomaisthorpe: Yeah that shouldn't have been there :)

internal/vulnerabilities/attackwave/lru_cache.go
```diff
@@ -0,0 +1,121 @@
...
+	return ent.value, true
+}
+
+func (c *lruCache) Set(key string, value int) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	now := time.Now()
+	expiry := time.Time{}
+	if c.ttl > 0 {
```
bitterpanda63: would this happen? - let's assume that there's always a ttl right, will make this lru cache easier to reason with

tomaisthorpe: Yeah, this is unnecessary, removed it!

bitterpanda63 (approved):

PR #196, opened 2025-11-28
Pin gotestsum version using Makefile instead of tools.go

tomaisthorpe:
Reducing amount of dependencies in the package, using the makefile to pin the version instead.


tomaisthorpe:
Changed my mind, went with a tools pkg instead:
- https://github.com/AikidoSec/firewall-go/pull/197

PR #206, opened 2025-12-03
Check for go files in compile args

tomaisthorpe:


cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -39,8 +39,25 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
...
 	}
 
+	// These are the arguments that we want to pass through to the compiler
+	// If we modify a file, we need to pass the modified file to the compiler instead of the original file
+	newArgs := make([]string, 0, len(toolArgs))
+
+	for _, arg := range toolArgs {
+		// If the argument starts with a dash or doesn't end with .go, it's not a Go file
+		// so we can pass it through to the compiler
+		if strings.HasPrefix(arg, "-") || !strings.HasSuffix(arg, ".go") {
```
bitterpanda63: is the !strings.HasSuffix .go not enough?

bitterpanda63 (approved):

cmd/zen-go/cmd_toolexec_compile.go
tomaisthorpe: It should be, removed the has prefix.

PR #207, opened 2025-12-03
Add wrap rules to instrumentor

tomaisthorpe:
Add some support for wrap rules, this is used for example adding middleware to Gin.

An example of running this locally on the `gin-postgres` sample app (requires a small modification to get things building):
```bash
go build -toolexec "$HOME/dev/firewall-go/cmd/zen-go/zen-go toolexec" -work .
```

Snippet of `go build` output with debug enabled:
```
# gin-postgres
zen-go: compiling package main
zen-go: transformed ./main.go -> $WORK/b001/zen-go/src/main.go
zen-go: added import zengin -> github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin
zen-go: processing import: key=zengin value=github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin
zen-go: adding to importcfg: github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin=/Users/tom/Library/Caches/go-build/99/99d90327049bfc90bbd0a77931377145352be9e9b936b5e811713b47cf6809e3-d
zen-go: wrote extended importcfg to $WORK/b001/zen-go/importcfg.txt
zen-go: replaced importcfg with $WORK/b001/zen-go/importcfg.txt
```

And result looks like:
```go
func main() {
	err := zen.Protect()
	if err != nil {
		log.Fatal(err)
	}

	db = NewDatabaseHelper()
	// Set up Gin router
	r := func() *gin.Engine {
		e := gin.
			Default()
		e.
			Use(zengin.
				GetMiddleware())
		return e
	}()
	
	// ...
}

### DIFF
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -5,12 +5,13 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
-	"github.com/urfave/cli/v3"
+	"github.com/AikidoSec/firewall-go/cmd/zen-go/internal"
 )
 
-func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer, tool string, toolArgs []string) error {
+func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, toolArgs []string) error {
 	// If has -V=full or just -V, then it's a version query
 	// Check for -V flag, with optional "full" value
 	for _, arg := range toolArgs {
@@ -34,14 +35,22 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
 		}
 	}
 
+	// Get objdir from output path (e.g., /tmp/go-build123/b001/_pkg_.a -> /tmp/go-build123/b001)
+	var objdir string
+	if outputPath != "" {
+		objdir = filepath.Dir(outputPath)
+	}
+
 	if isDebug() {
 		fmt.Fprintf(stderr, "zen-go: compiling package %s\n", pkgPath)
-		fmt.Fprintf(stderr, "%s, %s", importcfgPath, outputPath)
 	}
 
 	// These are the arguments that we want to pass through to the compiler
 	// If we modify a file, we need to pass the modified file to the compiler instead of the original file
 	newArgs := make([]string, 0, len(toolArgs))
+	allAddedImports := make(map[string]string) // import path -> alias
+
+	instrumentor := internal.NewInstrumentor()
 
 	for _, arg := range toolArgs {
 		// If doesn't end with .go, it's not a Go file
@@ -51,9 +60,63 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
 			continue
 		}
 
-		// It's a Go file, for now we just pass it through to the compiler
-		fmt.Fprintf(stderr, "zen-go: instrumenting file %s\n", arg)
-		newArgs = append(newArgs, arg)
+		result, modified, addedImports, err := instrumentor.InstrumentFile(arg, pkgPath)
+		if err != nil {
+			fmt.Fprintf(stderr, "zen-go: error instrumenting file %s: %v\n", arg, err)
+			newArgs = append(newArgs, arg)
+			continue
+		}
+
+		if !modified {
+			newArgs = append(newArgs, arg)
+			continue
+		}
+
+		// Track added imports
+		for alias, path := range addedImports {
+			allAddedImports[alias] = path
+		}
+
+		tempFile, err := writeTempFile(arg, result, objdir)
+		if err != nil {
+			fmt.Fprintf(stderr, "zen-go: error writing temp file: %v\n", err)
+			newArgs = append(newArgs, arg)
+			continue
+		}
+		newArgs = append(newArgs, tempFile)
+
+		if isDebug() {
+			fmt.Fprintf(stderr, "zen-go: transformed %s -> %s\n", arg, tempFile)
+			for alias, path := range allAddedImports {
+				fmt.Fprintf(stderr, "zen-go: added import %s -> %s\n", alias, path)
+			}
+		}
+	}
+
+	// If we added imports, we need to modify the importcfg
+	if len(allAddedImports) > 0 && importcfgPath != "" {
+		newImportcfg, err := internal.ExtendImportcfg(importcfgPath, allAddedImports, objdir, stderr, isDebug())
+		if err != nil {
+			fmt.Fprintf(stderr, "zen-go: warning: failed to extend importcfg: %v\n", err)
+		} else if newImportcfg != "" {
+			// Replace importcfg in args
+			for i := 0; i < len(newArgs); i++ {
+				if newArgs[i] == "-importcfg" && i+1 < len(newArgs) {
+					newArgs[i+1] = newImportcfg
+					if isDebug() {
+						fmt.Fprintf(stderr, "zen-go: replaced importcfg with %s\n", newImportcfg)
+					}
+					break
+				}
+				if strings.HasPrefix(newArgs[i], "-importcfg=") {
+					newArgs[i] = "-importcfg=" + newImportcfg
+					if isDebug() {
+						fmt.Fprintf(stderr, "zen-go: replaced importcfg with %s\n", newImportcfg)
+					}
+					break
+				}
+			}
+		}
 	}
 
 	// Run the compiler
@@ -84,4 +147,22 @@ func passthrough(tool string, args []string) error {
 	cmd.Stderr = os.Stderr
 
 	return cmd.Run()
+}
+
+func writeTempFile(origPath string, content []byte, objdir string) (string, error) {
+	// Write to objdir/zen-go/src/
+	dir := filepath.Join(objdir, "zen-go", "src")
+	if err := os.MkdirAll(dir, 0o755); err != nil {
+		// Fall back to source directory
+		dir = filepath.Dir(origPath)
+	}
+
+	base := filepath.Base(origPath)
+
+	// Use predictable name in objdir
+	outPath := filepath.Join(dir, base)
+	if err := os.WriteFile(outPath, content, 0o644); err != nil {
+		return "", err
+	}
+	return outPath, nil
 }
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -40,7 +40,8 @@ func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, too
 
 	// If we added imports, we need to modify the importcfg
 	if len(allAddedImports) > 0 && importcfgPath != "" {
-		if err := updateImportcfgInArgs(stderr, newArgs, importcfgPath, allAddedImports, objdir); err != nil {
+		newArgs, err = updateImportcfgInArgs(stderr, newArgs, importcfgPath, allAddedImports, objdir)
+		if err != nil {
 			fmt.Fprintf(stderr, "zen-go: warning: failed to update importcfg: %v\n", err)
 		}
 	}
@@ -152,35 +153,39 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 	return newArgs, allAddedImports, nil
 }
 
-func updateImportcfgInArgs(stderr io.Writer, args []string, importcfgPath string, addedImports map[string]string, objdir string) error {
+func updateImportcfgInArgs(stderr io.Writer, args []string, importcfgPath string, addedImports map[string]string, objdir string) ([]string, error) {
 	newImportcfg, err := internal.ExtendImportcfg(importcfgPath, addedImports, objdir, stderr, isDebug())
 	if err != nil {
-		return err
+		return args, err
 	}
+
 	if newImportcfg == "" {
-		return nil
+		return args, nil
 	}
 
-	replaceImportcfgArg(args, newImportcfg)
-
+	updatedArgs := replaceImportcfgArg(args, newImportcfg)
 	if isDebug() {
 		fmt.Fprintf(stderr, "zen-go: replaced importcfg with %s\n", newImportcfg)
 	}
 
-	return nil
+	return updatedArgs, nil
 }
 
-func replaceImportcfgArg(args []string, newPath string) {
-	for i := range len(args) {
-		if args[i] == "-importcfg" && i+1 < len(args) {
-			args[i+1] = newPath
-			return
+func replaceImportcfgArg(args []string, newPath string) []string {
+	result := make([]string, len(args))
+	copy(result, args)
+
+	for i := range len(result) {
+		if result[i] == "-importcfg" && i+1 < len(result) {
+			result[i+1] = newPath
+			return result
 		}
-		if strings.HasPrefix(args[i], "-importcfg=") {
-			args[i] = "-importcfg=" + newPath
-			return
+		if strings.HasPrefix(result[i], "-importcfg=") {
+			result[i] = "-importcfg=" + newPath
+			return result
 		}
 	}
+	return result
 }
 
 func writeTempFile(origPath string, content []byte, objdir string) (string, error) {
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -161,6 +161,7 @@ func writeTempFile(origPath string, content []byte, objdir string) (string, erro
 
 	// Use predictable name in objdir
 	outPath := filepath.Join(dir, base)
+	// #nosec G306 -- transformed source files need to be readable by the compiler
 	if err := os.WriteFile(outPath, content, 0o644); err != nil {
 		return "", err
 	}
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: cmd/zen-go/internal/instrumentor.go [hunk 1]
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type Instrumentor struct {
+	WrapRules []WrapRule
+}
+
+func NewInstrumentor() *Instrumentor {
+	return &Instrumentor{
+		WrapRules: []WrapRule{
+			{
```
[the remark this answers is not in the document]
tomaisthorpe: This will be replaced with loading rules from .yml files, just hardcoded here to keep this PR a bit more focused.

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,111 @@
...
+
+func transformDeclsWrap(decls []ast.Decl, fset *token.FileSet, pkgName, funcName string, rule WrapRule, modified *bool, importsToAdd map[string]string) {
+	for _, decl := range decls {
+		if fn, ok := decl.(*ast.FuncDecl); ok && fn.Body != nil {
+			transformStmtsWrap(fn.Body.List, fset, pkgName, funcName, rule, modified, importsToAdd)
+		}
+	}
+}
+
+func transformStmtsWrap(stmts []ast.Stmt, fset *token.FileSet, pkgName, funcName string, rule WrapRule, modified *bool, importsToAdd map[string]string) {
```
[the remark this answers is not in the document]
tomaisthorpe: Not a concern for this purpose, it will be compiling a variety of codebases but won't be running directly on live services, so no risk of DOS.

cmd/zen-go/internal/importcfg.go
```diff
@@ -0,0 +1,122 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+func ExtendImportcfg(origPath string, addedImports map[string]string, objdir string, stderr io.Writer, debug bool) (string, error) {
+	content, err := os.ReadFile(origPath)
```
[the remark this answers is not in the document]
bitterpanda63: locally run only

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -51,9 +60,63 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
...
+			newArgs = append(newArgs, arg)
+			continue
+		}
+
+		if !modified {
+			newArgs = append(newArgs, arg)
+			continue
+		}
+
+		// Track added imports
```
bitterpanda63: maybe instead of this comment give a quick example of what alias and path entails?

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -51,9 +60,63 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
...
+	}
+
+	// If we added imports, we need to modify the importcfg
+	if len(allAddedImports) > 0 && importcfgPath != "" {
+		newImportcfg, err := internal.ExtendImportcfg(importcfgPath, allAddedImports, objdir, stderr, isDebug())
+		if err != nil {
+			fmt.Fprintf(stderr, "zen-go: warning: failed to extend importcfg: %v\n", err)
+		} else if newImportcfg != "" {
+			// Replace importcfg in args
+			for i := 0; i < len(newArgs); i++ {
```
bitterpanda63: would maybe separate into function, is simple but long logic, function is already pretty long right?

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -85,3 +148,22 @@ func passthrough(tool string, args []string) error {
 
 	return cmd.Run()
 }
+
+func writeTempFile(origPath string, content []byte, objdir string) (string, error) {
+	// Write to objdir/zen-go/src/
+	dir := filepath.Join(objdir, "zen-go", "src")
```
bitterpanda63, line 193: mm what happens to these temporary files? where is the cleanup, or is that handled by go itself since we're in objdir?

cmd/zen-go/internal/instrumentor.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Add new imports
+	addImports(file, importsToAdd)
+
+	var out bytes.Buffer
+	if err := printer.Fprint(&out, fset, file); err != nil {
+		return nil, false, nil, err
+	}
+
+	return out.Bytes(), true, importsToAdd, nil
```
bitterpanda63: hard to read notation here, could we make this an object with clearly defined attributes?

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -51,9 +60,63 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
...
+		}
+
+		if !modified {
+			newArgs = append(newArgs, arg)
+			continue
+		}
+
+		// Track added imports
+		for alias, path := range addedImports {
+			allAddedImports[alias] = path
```
bitterpanda63: possible race condition to do this before the temp file write?

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -51,9 +60,63 @@ func toolexecCompileCommand(cmd *cli.Command, stdout io.Writer, stderr io.Writer
...
+				fmt.Fprintf(stderr, "zen-go: added import %s -> %s\n", alias, path)
+			}
+		}
+	}
+
+	// If we added imports, we need to modify the importcfg
+	if len(allAddedImports) > 0 && importcfgPath != "" {
+		newImportcfg, err := internal.ExtendImportcfg(importcfgPath, allAddedImports, objdir, stderr, isDebug())
+		if err != nil {
+			fmt.Fprintf(stderr, "zen-go: warning: failed to extend importcfg: %v\n", err)
```
bitterpanda63: revert changes made to files if failrue?

bitterpanda63: *args

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,111 @@
...
+					}
+				}
+			}
+		case *ast.BlockStmt:
+			transformStmtsWrap(s.List, fset, pkgName, funcName, rule, modified, importsToAdd)
+		}
+	}
+}
+
+func tryTransformCall(expr ast.Expr, fset *token.FileSet, pkgName, funcName string, rule WrapRule, modified *bool, importsToAdd map[string]string) ast.Expr {
```
bitterpanda63: this one def needs some comments explaining how we go about parsing & how the syntax tree works here I think. Or at least how we navigate it

cmd/zen-go/cmd_toolexec_compile.go
tomaisthorpe: If the project is built without the `-work` flag, then Go will clean everything up after it's done which is the normal way to run this. It's just creating a temporary folder.

If you build the project with `-work`, it will leave the files there, and then you can go see what has been instrumented.

cmd/zen-go/internal/instrumentor.go
tomaisthorpe: Added result struct 🫡

cmd/zen-go/internal/transform.go
tomaisthorpe: Added docs on how this works, was gonna do it later once I'd added the rest of the rules, but might as well add it now!

cmd/zen-go/cmd_toolexec_compile.go
tomaisthorpe: Split up 👍

cmd/zen-go/cmd_toolexec_compile.go
tomaisthorpe: For handling failures, I've added it to my list to properly look at, I think it makes it more complicated when we're messing with the linker, so it probably makes sense to attempt to look at that once that's in place.

cmd/zen-go/internal/instrumentor.go
```diff
@@ -0,0 +1,112 @@
...
+
+func (i *Instrumentor) InstrumentFile(filename string, compilingPkg string) (InstrumentFileResult, error) {
+	// Parse the file using go/parser
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, filename, nil, parser.ParseComments)
+	if err != nil {
+		return InstrumentFileResult{}, err
+	}
+
+	// Build map of existing imports
```
bitterpanda63, line 61:
```suggestion
	// Build map of existing imports to their local name
```

cmd/zen-go/internal/instrumentor.go
```diff
@@ -0,0 +1,112 @@
...
+	imports := make(map[string]string)
+	for _, imp := range file.Imports {
+		path, _ := strconv.Unquote(imp.Path.Value)
+		name := ""
+		if imp.Name != nil {
+			name = imp.Name.String()
+		} else {
+			parts := strings.Split(path, "/")
```
bitterpanda63, line 72: lets check why

cmd/zen-go/internal/instrumentor.go
```diff
@@ -0,0 +1,112 @@
...
+	modified := false
+	importsToAdd := make(map[string]string)
+
+	// Apply wrap rules
+	for _, rule := range i.WrapRules {
+		lastDot := strings.LastIndex(rule.MatchCall, ".")
+		if lastDot == -1 {
+			continue
+		}
+		pkg := rule.MatchCall[:lastDot]
```
bitterpanda63, line 91: maybe extract with an example

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,148 @@
...
+			if s.Body != nil {
+				for _, clause := range s.Body.List {
+					if cc, ok := clause.(*ast.CommClause); ok {
+						transformStmtsWrap(cc.Body, fset, pkgName, funcName, rule, modified, importsToAdd)
+					}
+				}
+			}
+		case *ast.BlockStmt:
+			transformStmtsWrap(s.List, fset, pkgName, funcName, rule, modified, importsToAdd)
+		}
```
bitterpanda63, line 125: check if its possible to have anonymous functions

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// Step 2: Check if the call is in the form "something.method()"
+	// (as opposed to just "function()")
+	sel, ok := call.Fun.(*ast.SelectorExpr)
+	if !ok {
+		return nil
+	}
+
+	// Step 3: Check if "something" is a simple identifier (package name)
```
bitterpanda63: Make sure both package name and function name match

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,148 @@
...
+	// This ensures we're dealing with pkgName.funcName, not x.y.z.funcName
+	ident, ok := sel.X.(*ast.Ident)
+	if !ok || ident.Name != pkgName || sel.Sel.Name != funcName {
+		return nil
+	}
+
+	// Step 4: Serialize the original call expression back to source code
+	// e.g., "http.Get(url)" becomes the string "http.Get(url)"
```
bitterpanda63:
```suggestion
	// e.g., (AST) "http.Get(url)" becomes the string "http.Get(url)"
```

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,148 @@
...
+	var origBuf bytes.Buffer
+	printer.Fprint(&origBuf, fset, call)
+	origCode := origBuf.String()
+
+	// Step 5: Apply the wrapping template
+	// e.g., if template is "instrument.Wrap({{.}})", we get "instrument.Wrap(http.Get(url))"
+	wrapped := strings.Replace(rule.WrapTmpl, "{{.}}", origCode, 1)
```
bitterpanda63, line 175: could we check if it was actually modified or not?

bitterpanda63: wrapped == origCode

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,148 @@
...
+
+	// Step 6: Parse the wrapped string back into an AST expression
+	wrappedExpr, err := parser.ParseExpr(wrapped)
+	if err != nil {
+		return nil
```
bitterpanda63: next pr

cmd/zen-go/internal/transform.go
tomaisthorpe: now returns an error if there's a parsing issue

cmd/zen-go/internal/transform.go
```diff
@@ -0,0 +1,169 @@
...
+					return err
+				}
+
+				if newExpr != nil {
+					s.Results[i] = newExpr
+				}
+			}
+		case *ast.IfStmt:
+			if s.Body != nil {
+				return transformStmtsWrap(s.Body.List, fset, pkgName, funcName, rule, modified, importsToAdd)
```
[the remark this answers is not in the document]
tomaisthorpe: This is okay, if there's an issue, we just want to stop.

cmd/zen-go/internal/transform.go
tomaisthorpe: Anonymous functions are not working correctly in some cases, it will require a few changes, would prefer to sort in a separate PR.

cmd/zen-go/internal/instrumentor.go
tomaisthorpe: Added comment explaining why this is here. Don't think I was fully awake when we went through it :)

cmd/zen-go/internal/imports.go
```diff
@@ -0,0 +1,49 @@
...
+	"go/token"
+	"strconv"
+)
+
+func addImports(f *ast.File, imports map[string]string) {
+	if len(imports) == 0 {
+		return
+	}
+
+	var importDecl *ast.GenDecl
```
bitterpanda63, line 24: trying to find first block of imports to add our ones to, if it doesnt exist yet we create our own

cmd/zen-go/internal/imports.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, spec := range importDecl.Specs {
+		if is, ok := spec.(*ast.ImportSpec); ok {
+			path, _ := strconv.Unquote(is.Path.Value)
+			existing[path] = true
+		}
+	}
+
+	for alias, path := range imports {
+		if existing[path] {
+			continue
```
bitterpanda63:
```suggestion
            // Making sure that the alias defined by the user for this path is not modified 
			continue
```

cmd/zen-go/internal/imports.go
```diff
@@ -0,0 +1,49 @@
...
+	if importDecl == nil {
+		importDecl = &ast.GenDecl{Tok: token.IMPORT}
+		f.Decls = append([]ast.Decl{importDecl}, f.Decls...)
+	}
+
+	existing := make(map[string]bool)
+	for _, spec := range importDecl.Specs {
+		if is, ok := spec.(*ast.ImportSpec); ok {
+			path, _ := strconv.Unquote(is.Path.Value)
+			existing[path] = true
```
bitterpanda63:
```suggestion
			existing[path] = alias
```

cmd/zen-go/internal/imports.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+	}
+
+	for alias, path := range imports {
+		if existing[path] {
```
bitterpanda63: if existing[path] && (existing[path].alias == undefined || existing[path].alias == alias)

cmd/zen-go/internal/imports.go
```diff
@@ -0,0 +1,49 @@
...
+			break
+		}
+	}
+
+	if importDecl == nil {
+		importDecl = &ast.GenDecl{Tok: token.IMPORT}
+		f.Decls = append([]ast.Decl{importDecl}, f.Decls...)
+	}
+
+	existing := make(map[string]bool)
```
bitterpanda63: create an array of already defined aliases

bitterpanda63: db myssql
db pg

cmd/zen-go/internal/importcfg.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	// Parse existing entries
+	existing := make(map[string]bool)
+	for _, line := range strings.Split(string(content), "\n") {
+		if strings.HasPrefix(line, "packagefile ") {
+			parts := strings.SplitN(strings.TrimPrefix(line, "packagefile "), "=", 2)
```
bitterpanda63, line 22: SplitN -> Split

cmd/zen-go/internal/importcfg.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func createTempFile(objdir string) (*os.File, error) {
+	if objdir != "" {
+		dir := filepath.Join(objdir, "zen-go")
+		if err := os.MkdirAll(dir, 0o755); err == nil {
+			return os.Create(filepath.Join(dir, "importcfg.txt"))
+		}
+	}
+	return os.CreateTemp("", "importcfg_*.txt")
```
bitterpanda63, line 90: bit weird

bitterpanda63: is go notation

bitterpanda63 (approved):

PR #209, opened 2025-12-03
Add prepend rules to instrumentor

tomaisthorpe:


cmd/zen-go/internal/transform.go
```diff
@@ -188,3 +190,114 @@ func tryTransformCall(expr ast.Expr, fset *token.FileSet, pkgName, funcName stri
 
 	return wrappedExpr, nil
 }
+
+// transformDeclsPrepend finds function declarations matching the prepend rule
+// and prepends statements to the function body
+func transformDeclsPrepend(decls []ast.Decl, compilingPkg string, rule PrependRule, modified *bool, importsToAdd map[string]string) error {
```
bitterpanda63: maybe would try and keep a bit seperated, this transform.go is otherwise going to become fairly big? up to you

cmd/zen-go/internal/transform.go
```diff
@@ -188,3 +190,114 @@ func tryTransformCall(expr ast.Expr, fset *token.FileSet, pkgName, funcName stri
...
+	for _, decl := range decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Body == nil || fn.Recv == nil {
+			continue
+		}
+
+		// Check if function name matches
+		if fn.Name.Name != rule.FuncName {
```
bitterpanda63: should we also not check package name or am I missing something?

cmd/zen-go/internal/transform.go
```diff
@@ -188,3 +190,114 @@ func tryTransformCall(expr ast.Expr, fset *token.FileSet, pkgName, funcName stri
...
+		return false
+	}
+
+	recvField := recv.List[0]
+	recvType := formatReceiverType(recvField.Type, compilingPkg)
+
+	return recvType == expectedType
+}
+
+// formatReceiverType formats a receiver type expression as a string
```
bitterpanda63: can you give some examples here, assuming this is to match with wrapper rules right?

cmd/zen-go/internal/transform.go
```diff
@@ -188,3 +190,114 @@ func tryTransformCall(expr ast.Expr, fset *token.FileSet, pkgName, funcName stri
...
+		if ident, ok := t.X.(*ast.Ident); ok {
+			return ident.Name + "." + t.Sel.Name
+		}
+	}
+	return ""
+}
+
+// substituteArguments replaces {{ .Function.Argument N }} with actual parameter names
+func substituteArguments(tmpl string, fn *ast.FuncDecl) (string, error) {
+	t, err := template.New("tmpl").Parse(tmpl)
```
bitterpanda63: why are we using html/template here but then we have our own template, its kinda confusing

cmd/zen-go/internal/transform.go
tomaisthorpe: We check the receiver type which is the next condition, that's what we care about, it's not a package we check for prepend rules.

cmd/zen-go/internal/transform.go
tomaisthorpe: I think it's fine at the moment, but will get better with the linker work, so just separated them.

cmd/zen-go/internal/transform.go
tomaisthorpe: It should have been `text/template` which I've corrected, but I'm not quite following the question, we're parsing the templates here so that we can inject any values we require.

bitterpanda63 (approved):

PR #215, opened 2025-12-04
Return rate limit data directly instead of key

tomaisthorpe:
Unnecessary two step operation to get rate limiting config instead of just returning it directly.

### DIFF
--- a/internal/agent/ratelimiting/rate_limiting.go
+++ b/internal/agent/ratelimiting/rate_limiting.go
@@ -123,13 +123,8 @@ func (rl *RateLimiter) checkEntity(method string, route string, kind entityKind,
 	rl.mu.Lock()
 	defer rl.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


internal/agent/ratelimiting/rate_limiting.go
```diff
@@ -203,7 +198,7 @@ func (rl *RateLimiter) findMatchingRateLimitEndpoint(method string, route string
 		rate := float64(data.Config.MaxRequests) / float64(data.Config.WindowSizeInMS)
 
 		if mostRestrictive == nil || rate < lowestRate {
-			mostRestrictive = &key
+			mostRestrictive = data
```
bitterpanda63: now that the data is used multiple places in this for loop, its name is kinda vague? maybe currentEndpointData?

bitterpanda63 (approved):

PR #221, opened 2025-12-10
Add pgx native driver support

tomaisthorpe:


sample-apps/echo-postgres/go.sum
bitterpanda63, line 1: should echo-postgres be changed?

instrumentation/sinks/jackc/pgx/orchestrion.yml
```diff
@@ -0,0 +1,41 @@
...
+          - receiver: "*github.com/jackc/pgx/v5.Conn"
+          - name: Query
+    advice:
+      - prepend-statements:
+          imports:
+            sink: github.com/AikidoSec/firewall-go/instrumentation/sinks/jackc/pgx
+          template: |-
+            _aikido_block := sink.ExamineContext({{ .Function.Argument 0}}, {{ .Function.Argument 1}}, "pgx.Conn.Query")
+            if _aikido_block != nil {
+              return &baseRows{err: _aikido_block}, _aikido_block
```
bitterpanda63, line 22: do we have to create these baseRows, we can't just pass along nil? bit confused why there are 2 errors returned this way

instrumentation/sinks/jackc/pgx/orchestrion.yml
```diff
@@ -0,0 +1,41 @@
...
+          - name: Exec
+    advice:
+      - prepend-statements:
+          imports:
+            sink: github.com/AikidoSec/firewall-go/instrumentation/sinks/jackc/pgx
+          template: |-
+            _aikido_block := sink.ExamineContext({{ .Function.Argument 0}}, {{ .Function.Argument 1}}, "pgx.Conn.Exec")
+            if _aikido_block != nil {
+              var _zeroTag pgconn.CommandTag
+              return _zeroTag, _aikido_block
```
bitterpanda63, line 40: why not here as well just return nil, _aikido_block?

instrumentation/sinks/jackc/pgx/examine.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/AikidoSec/firewall-go/zen"
+)
+
+// ExamineContext checks for SQL injection vulnerabilities in database queries.
+// This function is called by the instrumentation framework to scan SQL queries
+// before they are executed against the database.
+func ExamineContext(ctx context.Context, query string, op string) error {
+	err := vulnerabilities.Scan(ctx, op, sqlinjection.SQLInjectionVulnerability, &sqlinjection.ScanArgs{
+		Statement: query,
+		Dialect:   "default",
```
bitterpanda63: we can pass along its a postgres dialect no?

sample-apps/echo-postgres/go.sum
tomaisthorpe: Currently yes, this is why I want to split up the instrumentation packages. Otherwise you get packages in apps that aren't actually used.

instrumentation/sinks/jackc/pgx/orchestrion.yml
tomaisthorpe: This is just how `pgx` works, in some methods it ignores the error and assumes the error is also in `rows`:
https://github.com/jackc/pgx/blob/a2fca037434a0a7096b095d4ed87cdffb03b626e/conn.go#L904C15-L904C20

instrumentation/sinks/jackc/pgx/orchestrion.yml
tomaisthorpe: It's not a pointer or an interface, so we can't return nil. It's the same pattern we see in pgx itself:
https://github.com/jackc/pgx/blob/a2fca037434a0a7096b095d4ed87cdffb03b626e/conn.go#L463

instrumentation/sinks/jackc/pgx/orchestrion.yml
```diff
@@ -0,0 +1,41 @@
...
+            }
```
bitterpanda63:
```suggestion
            }

```

bitterpanda63 (approved):

PR #223, opened 2025-12-10
Remove URL from RouteMetadata

tomaisthorpe:
The full URL isn't used as part of the route matching, so removed it.


hansott (approved):
Assuming we match the wildcards against the route pattern, instead of actual URL (https://github.com/AikidoSec/firewall-node/blob/main/library/helpers/matchEndpoints.ts#L44), which in most cases seems fine indeed.

PR #245, opened 2025-12-18
Add support for inject declarations

tomaisthorpe:
## Summary by Aikido
| ⚠️ Security Issues: 2 | 🔍 Quality Issues: 5 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added InjectDeclRule and AST transform to inject declarations before anchors.
* Introduced link-deps sidecar files and read/write helpers for archives.

**⚡ Enhancements**
* Integrated compile and link to record and resolve link-time dependencies.

**🔧 Refactors**
* Refactored Instrumentor to accept InstrumentationRules and return link metadata.

**📚 Documentation**
* Added net/http instrumentation YAML and comments explaining linkname usage.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/75261604?groupId=6)</sup>

### DIFF
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -33,7 +33,7 @@ func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, too
 
 	// These are the arguments that we want to pass through to the compiler
 	// If we modify a file, we need to pass the modified file to the compiler instead of the original file
-	newArgs, allAddedImports, err := instrumentFiles(stderr, toolArgs, pkgPath, objdir)
+	newArgs, allAddedImports, allLinkDeps, err := instrumentFiles(stderr, toolArgs, pkgPath, objdir)
 	if err != nil {
 		return err
 	}
@@ -50,6 +50,12 @@ func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, too
 	err = passthrough(tool, newArgs)
 	if err != nil {
 		return err
+	}
+
+	// After successful compile, if we have link-time dependencies, record them
+	// in a sidecar file next to the archive. The linker will read these later.
+	if len(allLinkDeps) > 0 {
+		writeLinkDepsForArchive(stderr, outputPath, allLinkDeps)
 	}
 
 	return nil
@@ -105,12 +111,13 @@ func getObjDir(outputPath string) string {
 	return filepath.Dir(outputPath)
 }
 
-func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string) ([]string, map[string]string, error) {
+func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string) ([]string, map[string]string, []string, error) {
 	newArgs := make([]string, 0, len(toolArgs))
-	allAddedImports := make(map[string]string) // import path -> alias
+	allAddedImports := make(map[string]string) // alias -> import path
+	var allLinkDeps []string
 	instrumentor, err := internal.NewInstrumentor()
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
 	for _, arg := range toolArgs {
@@ -137,6 +144,9 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 		// zengin -> github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin
 		maps.Copy(allAddedImports, result.Imports)
 
+		// Collect link dependencies
+		allLinkDeps = append(allLinkDeps, result.LinkDeps...)
+
 		tempFile, err := writeTempFile(arg, result.Code, objdir)
 		if err != nil {
 			fmt.Fprintf(stderr, "zen-go: error writing temp file: %v\n", err)
@@ -147,13 +157,16 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 
 		if isDebug() {
 			fmt.Fprintf(stderr, "zen-go: transformed %s -> %s\n", arg, tempFile)
-			for alias, path := range allAddedImports {
+			for alias, path := range result.Imports {
 				fmt.Fprintf(stderr, "zen-go: added import %s -> %s\n", alias, path)
+			}
+			for _, dep := range result.LinkDeps {
+				fmt.Fprintf(stderr, "zen-go: added link dep %s\n", dep)
 			}
 		}
 	}
 
-	return newArgs, allAddedImports, nil
+	return newArgs, allAddedImports, allLinkDeps, nil
 }
 
 func updateImportcfgInArgs(stderr io.Writer, args []string, importcfgPath string, addedImports map[string]string, objdir string) ([]string, error) {
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -54,9 +54,7 @@ func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, too
 
 	// After successful compile, if we have link-time dependencies, record them
 	// in a sidecar file next to the archive. The linker will read these later.
-	if len(allLinkDeps) > 0 {
-		writeLinkDepsForArchive(stderr, outputPath, allLinkDeps)
-	}
+	writeLinkDepsForArchive(stderr, outputPath, allLinkDeps)
 
 	return nil
 }
--- a/cmd/zen-go/cmd_toolexec_compile.go
+++ b/cmd/zen-go/cmd_toolexec_compile.go
@@ -114,7 +114,7 @@ func getObjDir(outputPath string) string {
 func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string) ([]string, map[string]string, []string, error) {
 	newArgs := make([]string, 0, len(toolArgs))
 	allAddedImports := make(map[string]string) // alias -> import path
-	var allLinkDeps []string
+	var allAddedLinkDeps []string
 	instrumentor, err := internal.NewInstrumentor()
 	if err != nil {
 		return nil, nil, nil, err
@@ -144,8 +144,8 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 		// zengin -> github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin
 		maps.Copy(allAddedImports, result.Imports)
 
-		// Collect link dependencies
-		allLinkDeps = append(allLinkDeps, result.LinkDeps...)
+		// Collect link dependencies from each file to pass to the linker later
+		allAddedLinkDeps = append(allAddedLinkDeps, result.LinkDeps...)
 
 		tempFile, err := writeTempFile(arg, result.Code, objdir)
 		if err != nil {
@@ -166,7 +166,7 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 		}
 	}
 
-	return newArgs, allAddedImports, allLinkDeps, nil
+	return newArgs, allAddedImports, allAddedLinkDeps, nil
 }
 
 func updateImportcfgInArgs(stderr io.Writer, args []string, importcfgPath string, addedImports map[string]string, objdir string) ([]string, error) {
--- a/cmd/zen-go/cmd_toolexec_link.go
+++ b/cmd/zen-go/cmd_toolexec_link.go
@@ -25,6 +25,7 @@ func toolexecLinkCommand(_ io.Writer, stderr io.Writer, tool string, toolArgs []
 	}
 
 	// Read the importcfg to find all archives
+	// #nosec G304 - importcfgPath comes from Go toolchain args, not user input
 	content, err := os.ReadFile(importcfgPath)
 	if err != nil {
 		return passthrough(tool, toolArgs)
@@ -51,7 +52,7 @@ func toolexecLinkCommand(_ io.Writer, stderr io.Writer, tool string, toolArgs []
 		fmt.Fprintf(stderr, "zen-go: warning: failed to write extended importcfg: %v\n", err)
 		return passthrough(tool, toolArgs)
 	}
-	defer os.Remove(newImportcfgPath)
+	defer func() { _ = os.Remove(newImportcfgPath) }()
 
 	// Update args with new importcfg
 	newArgs := replaceLinkerImportcfgArg(toolArgs, newImportcfgPath)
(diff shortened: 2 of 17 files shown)


cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		archivePath := parts[1]
+		deps, err := internal.ReadLinkDeps(archivePath)
+		if err != nil {
+			continue
+		}
+
+		if len(deps) > 0 {
+			for _, dep := range deps {
```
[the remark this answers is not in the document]
tomaisthorpe: I don't think it's necessary for this function.

cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,195 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/AikidoSec/firewall-go/cmd/zen-go/internal"
+)
+
+func toolexecLinkCommand(_ io.Writer, stderr io.Writer, tool string, toolArgs []string) error {
```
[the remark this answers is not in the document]
tomaisthorpe: Everything is already split into smaller functions, don't think adds value splitting further.

cmd/zen-go/internal/rules.go
```diff
@@ -29,31 +29,51 @@ type Rule struct {
...
 type InstrumentationRules struct {
-	WrapRules    []WrapRule
-	PrependRules []PrependRule
+	WrapRules       []WrapRule
+	PrependRules    []PrependRule
+	InjectDeclRules []InjectDeclRule
+}
+
+// WrapRule wraps a function call expression
+type WrapRule struct {
```
bitterpanda63, line 49: weird, where was this WrapRule defined before?

tomaisthorpe: It was in `instrumentor.go`, so moved here to make it all consistent.

cmd/zen-go/internal/instrumentor.go
```diff
@@ -31,22 +24,31 @@ func NewInstrumentor() (*Instrumentor, error) {
 			return nil, err
 		}
 
-		return &Instrumentor{WrapRules: rules.WrapRules, PrependRules: rules.PrependRules}, nil
+		return &Instrumentor{
```
bitterpanda63: is same with passing along to `NewInstrumentorWithRules` right?, What is the use of that function

bitterpanda63: looks like it's only used in test cases, but Instruemntor itself is not internal rgiht, so see no use in defining this here?

cmd/zen-go/internal/transform_inject_decl.go
```diff
@@ -0,0 +1,120 @@
...
+// The declaration is inserted immediately before the anchor function.
+func transformDeclsInjectDecl(f *ast.File, fset *token.FileSet, rule InjectDeclRule, modified *bool, linksToAdd *[]string) error {
+	anchorIndex := findAnchorFunction(f, rule.AnchorFunc)
+	if anchorIndex == -1 {
+		return nil // Anchor function not found, skip
+	}
+
+	// Parse the declaration template with comments
+	// We need to add "unsafe" import for go:linkname to work
+	declCode := "package p\nimport _ \"unsafe\"\n" + rule.DeclTemplate
```
bitterpanda63: this is pretty hard to read with the newlines and `"` all in there

cmd/zen-go/internal/hash.go
```diff
@@ -27,6 +27,41 @@ func ComputeInstrumentationHash(inst *Instrumentor) string {
 		fmt.Fprintf(h, "%s\n", rule.WrapTmpl)
 	}
 
+	// Hash prepend rules
+	for _, rule := range inst.PrependRules {
```
bitterpanda63, line 31: mm, I def should've noticed this was absent in the prepend pr my bad

bitterpanda63: can we extract to its own pr though, this is independent still right

cmd/zen-go/internal/importcfg.go
```diff
@@ -90,7 +90,7 @@ func createTempFile(objdir string) (*os.File, error) {
 	return os.CreateTemp("", "importcfg_*.txt")
 }
 
-func getPackageExport(importPath string) (string, error) {
+func GetPackageExport(importPath string) (string, error) {
```
bitterpanda63, line 103: docstring

bitterpanda63: importPath = cmd/zen-go/internal
return = (<path on disk>, err)

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -137,6 +144,9 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 		// zengin -> github.com/AikidoSec/firewall-go/instrumentation/sources/gin-gonic/gin
 		maps.Copy(allAddedImports, result.Imports)
 
+		// Collect link dependencies
```
bitterpanda63: give a clearer comment, crucial

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -105,12 +111,13 @@ func getObjDir(outputPath string) string {
 	return filepath.Dir(outputPath)
 }
 
-func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string) ([]string, map[string]string, error) {
+func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string) ([]string, map[string]string, []string, error) {
 	newArgs := make([]string, 0, len(toolArgs))
-	allAddedImports := make(map[string]string) // import path -> alias
+	allAddedImports := make(map[string]string) // alias -> import path
+	var allLinkDeps []string
```
bitterpanda63: allAddedLinkDeps -> ours
allLinkDeps -> existing + ours

cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,196 @@
...
```
bitterpanda63, line 17: add example of how this link command looks?

bitterpanda63: link to go documentation

cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,196 @@
...
+		fmt.Fprintf(stderr, "zen-go: intercepting linker\n")
+	}
+
+	// Read the importcfg to find all archives
+	content, err := os.ReadFile(importcfgPath)
+	if err != nil {
+		return passthrough(tool, toolArgs)
+	}
+
+	// Collect all link deps from all archives
```
bitterpanda63, line 34: archives -> sidecars?

bitterpanda63: (temporary files between compile and linker stage)

cmd/zen-go/internal/linkdeps.go
```diff
@@ -0,0 +1,49 @@
...
+		fmt.Fprintf(stderr, "zen-go: wrote link deps to %s: %v\n", depsFile, deps)
+	}
+
+	return nil
+}
+
+// ReadLinkDeps reads link-time dependencies from a sidecar file next to the archive.
+// Returns nil if the file doesn't exist (which is not an error - it means no link deps).
+func ReadLinkDeps(archivePath string) ([]string, error) {
+	depsFile := archivePath + ".linkdeps"
```
bitterpanda63: all called archive still?

cmd/zen-go/internal/linkdeps.go
```diff
@@ -0,0 +1,49 @@
...
+	"io"
+	"os"
+	"strings"
+)
+
+// WriteLinkDeps writes link-time dependencies to a sidecar file next to the archive.
+// These dependencies are packages that need to be linked into the final binary
+// but are not direct imports of the package being compiled.
+func WriteLinkDeps(archivePath string, deps []string, stderr io.Writer, debug bool) error {
+	depsFile := archivePath + ".linkdeps"
```
bitterpanda63: feels a bit dangerous to have that ".linkdeps" twice

cmd/zen-go/internal/hash.go
tomaisthorpe: https://github.com/AikidoSec/firewall-go/pull/260

cmd/zen-go/cmd_toolexec_link.go
tomaisthorpe: archives is the correct term here, the sidecars are next to`.a` files after they've been compiled.

cmd/zen-go/internal/instrumentor.go
tomaisthorpe: I think it's best keeping `NewInstrumentorWithRules`, but I've updated `NewInstrumentor` to use it. It makes it much simpler for the tests, and doesn't really add any extra complication having it here.

cmd/zen-go/cmd_toolexec_link.go
bitterpanda63: yep idd, some confusion on my end

cmd/zen-go/internal/instrumentor.go
bitterpanda63: also works ye

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -147,13 +157,16 @@ func instrumentFiles(stderr io.Writer, toolArgs []string, pkgPath, objdir string
 
 		if isDebug() {
 			fmt.Fprintf(stderr, "zen-go: transformed %s -> %s\n", arg, tempFile)
-			for alias, path := range allAddedImports {
+			for alias, path := range result.Imports {
```
bitterpanda63, line 158: it was printing all added imports in each iteration? result.* is only for the current iteration then

cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,201 @@
...
+			result[i] = "-importcfg=" + newPath
+			return result
+		}
+	}
+	return result
+}
+
+// writeLinkDepsForArchive writes link dependencies for the compiled archive if any exist.
+func writeLinkDepsForArchive(stderr io.Writer, outputPath string, linkDeps []string) {
+	if len(linkDeps) == 0 {
```
bitterpanda63, line 189: we do this check twice?

cmd/zen-go/cmd_toolexec_compile.go
```diff
@@ -52,6 +52,12 @@ func toolexecCompileCommand(stdout io.Writer, stderr io.Writer, tool string, too
 		return err
 	}
 
+	// After successful compile, if we have link-time dependencies, record them
+	// in a sidecar file next to the archive. The linker will read these later.
+	if len(allLinkDeps) > 0 {
```
bitterpanda63: just pass it along, the length check already happens inside the function

cmd/zen-go/internal/transform_inject_decl.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, decl := range declFile.Decls {
+		// Skip the import declaration, we'll handle that separately
+		if genDecl, ok := decl.(*ast.GenDecl); ok && genDecl.Tok == token.IMPORT {
+			continue
+		}
+		declsToInject = append(declsToInject, decl)
+	}
+
+	if len(declsToInject) == 0 {
+		return nil
```
bitterpanda63: feels more like an error even if there arent any decl, there's no way that the only declaration being found after parsing is an import one right 🤔 - then there would be no function def?

cmd/zen-go/cmd_toolexec_link.go
```diff
@@ -0,0 +1,201 @@
...
+		return passthrough(tool, toolArgs)
+	}
+
+	// Collect all link deps from all archives
+	allLinkDeps := collectLinkDeps(content, stderr)
+	if len(allLinkDeps) == 0 {
+		return passthrough(tool, toolArgs)
+	}
+
+	// Check which deps are missing from importcfg and get their export paths
```
bitterpanda63: its a bit of a vague comment, we get the linked deps with collectLinkDeps from the importcfg already, what would be missing here exactly? Would the archive path be there but not the export path? could we add some examples possibly?

bitterpanda63: or are these our new deps?

cmd/zen-go/internal/transform_inject_decl.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	// Find and associate comments with declarations
+	// Comments that appear immediately before a declaration should be preserved
```
bitterpanda63: whats the reason for this? just clarity during debugging (assuming these are the comments before function like _os_examine once resolved? or is it the go:linkname comment?

cmd/zen-go/cmd_toolexec_compile.go
tomaisthorpe: yeah exactly, was a mistake in the original implementation

cmd/zen-go/internal/transform_inject_decl.go
tomaisthorpe: think that's a fair comment, we probably want it to force it to fail if there's no declaration in the template. if that's the case, we did something wrong writing the yaml and want to catch it

cmd/zen-go/internal/transform_inject_decl.go
tomaisthorpe: It's for preserving the go:linkname compiler directive. Go's AST stores comments separately from declarations, so when we copy the function declaration to the target file, we need to explicitly copy the associated //go:linkname comment too, otherwise the directive would be lost and linking would fail. I'll update the comment to make it clearer.

cmd/zen-go/cmd_toolexec_link.go
tomaisthorpe: added a better comment!

cmd/zen-go/internal/transform_inject_decl.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	// Copy go:linkname directive comments to the target file.
+	// Go's AST stores comments separately from declarations, so we must
+	// explicitly add them to f.Comments or they'll be lost in the output.
+	for _, decl := range declsToInject {
+		for _, cg := range declFile.Comments {
+			// Check if comment is immediately before this declaration
+			if decl.Pos() > cg.Pos() && decl.Pos()-cg.End() < 100 {
+				f.Comments = append(f.Comments, cg)
```
bitterpanda63, line 51: this doesnt mess with the anchor index because a comment != declaration I asusme?

cmd/zen-go/internal/transform_inject_decl.go
```diff
@@ -0,0 +1,125 @@
...
+	// Find or create an import declaration
+	var importDecl *ast.GenDecl
+	for _, decl := range f.Decls {
+		if genDecl, ok := decl.(*ast.GenDecl); ok && genDecl.Tok == token.IMPORT {
+			importDecl = genDecl
+			break
+		}
+	}
+
+	unsafeSpec := &ast.ImportSpec{
```
bitterpanda63, line 107: does the unsafe import not mess with type checking in the compiler that would oterhwise still be performed?

bitterpanda63 (approved):

PR #247, opened 2025-12-19
Add support for chi

tomaisthorpe:
## Summary by Aikido
**🚀 New Features**
* Added go-chi/chi instrumentation with middleware and request parsing.

**⚡ Enhancements**
* Registered chi source and enabled chi sample app in e2e tests.


instrumentation/sources/labstack/echo/orchestrion.yml
```diff
@@ -1,8 +1,3 @@
-# Unless explicitly stated otherwise all files in this repository are licensed
```
bitterpanda63, line 1: add into separate PR?

instrumentation/sources/go-chi/chi/middleware.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/AikidoSec/firewall-go/internal/log"
+	"github.com/AikidoSec/firewall-go/internal/request"
+	"github.com/go-chi/chi/v5"
+	"github.com/go-chi/chi/v5/middleware"
+)
+
+// GetMiddleware returns middleware that will create contexts of incoming requests.
+func GetMiddleware() func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
bitterpanda63: I sort of agree here, not that it's complex, but that its a bit hard to have 3 func def inside each other (I get why we do it, but we could also put the `func(w http.ResponseWriter, r *http.Request) {` under it?

instrumentation/sources/go-chi/chi/middleware.go
```diff
@@ -0,0 +1,116 @@
...
+			var routeParams map[string]string
+
+			if routeCtx != nil {
+				var urlParams chi.RouteParams
+				route, urlParams = getRoutePattern(r)
+				if len(urlParams.Keys) > 0 {
+					routeParams = make(map[string]string, len(urlParams.Keys))
+					for i, key := range urlParams.Keys {
+						if i < len(urlParams.Values) {
+							routeParams[key] = urlParams.Values[i]
```
bitterpanda63, line 44: are these query parameters? If so, no possibility for e.g. having the same key with different values? 
`?a=1&a=2&a=3`

bitterpanda63: nvm, okey so guessing query params are passed along wiht "Route" itself?

bitterpanda63 (approved):

PR #248, opened 2025-12-19
Add instance pool with max age for instances

tomaisthorpe:
The WASM instances seem to suffer from memory fragmentation over time resulting in a very slow leak. Instances now stick around for 5 minutes before being dropped.

## Summary by Aikido
**🚀 New Features**
* Introduced a custom Pool type managing WASM instances and lifecycle.

**⚡ Enhancements**
* Evicted WASM instances older than five minutes and capped idle.

**🐛 Bugfixes**
* Added nil checks and error handling when closing or acquiring instances.

**🔧 Refactors**
* Replaced sync.Pool with custom Pool and updated instance creation signatures.

### DIFF
--- a/internal/vulnerabilities/zeninternals/pool.go
+++ b/internal/vulnerabilities/zeninternals/pool.go
@@ -54,7 +54,10 @@ func (p *Pool) tryGetInstance() *wasmInstance {
 
 		// Check age of instance, if it's too old, we should discard it
 		if instance.createdAt.Before(time.Now().Add(-instanceMaxAge)) {
-			instance.mod.Close(context.Background())
+			if instance.mod != nil {
+				instance.mod.Close(context.Background())
+			}
+
 			return nil
 		}
 
--- a/internal/vulnerabilities/zeninternals/pool.go
+++ b/internal/vulnerabilities/zeninternals/pool.go
@@ -1,0 +1,67 @@
+package zeninternals
+
+import (
+	"errors"
+	"sync"
+	"time"
+)
+
+const (
+	instanceMaxAge   = 5 * time.Minute
+	maxIdleInstances = 10
+)
+
+type Pool struct {
+	newInstance   func() *wasmInstance
+	idleInstances []*wasmInstance
+
+	mu sync.Mutex
+}
+
+func NewPool(newInstance func() *wasmInstance) *Pool {
+	return &Pool{
+		newInstance:   newInstance,
+		idleInstances: make([]*wasmInstance, 0),
+	}
+}
+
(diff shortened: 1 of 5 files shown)


internal/vulnerabilities/zeninternals/pool.go
```diff
@@ -0,0 +1,67 @@
...
+		idleInstances: make([]*wasmInstance, 0),
+	}
+}
+
+func (p *Pool) Get() (*wasmInstance, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// Loop through max idle + 1 to ensure run through all available instances
+	for range maxIdleInstances + 1 {
```
bitterpanda63, line 35: not sure how there could be 11 in here exactly? It's limited at 9 right?

internal/vulnerabilities/zeninternals/pool.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (p *Pool) tryGetInstance() *wasmInstance {
+	if len(p.idleInstances) > 0 {
+		instance := p.idleInstances[len(p.idleInstances)-1]
+		p.idleInstances = p.idleInstances[:len(p.idleInstances)-1]
+
+		// Check health of instance
+		if instance.createdAt.Before(time.Now().Add(-instanceMaxAge)) {
+			return nil
```
bitterpanda63: wouldn't we want to get another instance, or is this just internal and is `Get` the only one we will really use? (mb add comment explaining)

internal/vulnerabilities/zeninternals/pool.go
tomaisthorpe: I've adjusted the comment slightly. The idea is that if all instances are too old, we should loop through all of them, which will return nil, and then we run once more to get a fresh instance, in reality, it's pretty rate that we would loop this many times, but this is a good upper limit to ensure we get an instance back.

internal/vulnerabilities/zeninternals/pool.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func (p *Pool) Get() (*wasmInstance, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// Loop through max idle + 1 to ensure we can loop through all potentially expired instances,
+	// and still get a new instance.
+	for range maxIdleInstances + 1 {
```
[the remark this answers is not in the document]
tomaisthorpe: false positive, for range over ints are valid

bitterpanda63 (approved):

PR #261, opened 2026-01-07
Switch back to transits for some of the inject declaration instrumentation

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added transits API to centralize path and command examination.

**⚡ Enhancements**
* Replaced inject declarations in instrumentation YAMLs with transit calls.
* Registered transits during Protect startup by calling SetupTransits.

**🔧 Refactors**
* Moved path and command scanning logic into internal transits implementation.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/73056297?groupId=6)</sup>


instrumentation/sinks/path/orchestrion.yml
```diff
@@ -12,13 +12,10 @@ aspects:
...
-          template: |-
-            //go:linkname __aikido_path_Examine github.com/AikidoSec/firewall-go/instrumentation/sinks/path.Examine
-            func __aikido_path_Examine ([]string) error
       - prepend-statements:
+          imports:
+            zentransits: github.com/AikidoSec/firewall-go/internal/vulnerabilities/transits
           template: |-
-            __aikido_path_Examine({{ .Function.Argument 0 }})
+            if examinePath := zentransits.ExaminePath(); examinePath != nil {
+              examinePath("path.Join", {{ .Function.Argument 0 }}, true)
```
bitterpanda63: now this true is very vague, its the deferred reporting right?

instrumentation/sinks/os/orchestrion.yml
```diff
@@ -39,8 +21,12 @@ aspects:
...
     advice:
       - prepend-statements:
+          imports:
+            zentransits: github.com/AikidoSec/firewall-go/internal/vulnerabilities/transits
           template: |-
-            _aikido_block := __aikido_os_Examine({{ .Function.Argument 0 }})
-            if _aikido_block != nil {
-              return nil, _aikido_block
+            if examinePath := zentransits.ExaminePath(); examinePath != nil {
+              _aikido_block := examinePath("os.OpenFile", []string{{`{`}}{{ .Function.Argument 0 }}{{`}`}}, false)
```
bitterpanda63: the `{` alone is setting of orchestrion? damn. not the case with the Template stuff we use right? or is this just a go thing

tomaisthorpe: It's the go templating, it's looking for `{{`, and we have `{{{` because of the string slice. This is just how we need to escape it.

instrumentation/sinks/path/orchestrion.yml
tomaisthorpe: I've added a const to make it a bit clearer what this is for.

bitterpanda63 (approved):

PR #269, opened 2026-01-14
Use packages.Load instead of calling "go list" directly

tomaisthorpe:
## Summary by Aikido
|  Security Issues: 0 |  Quality Issues: 0 | ✅ Resolved Issues: 1 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Removed os/exec dependency and simplified module-root discovery logic
* Added error when package was not resolved during packages.Load

**🔧 Refactors**
* Replaced direct 'go list' call with packages.Load usage

<sup>[More info](https://app.aikido.dev/featurebranch/scan/74673840?groupId=6)</sup>


cmd/zen-go/internal/importcfg.go
```diff
@@ -91,16 +93,30 @@ func createTempFile(objdir string) (*os.File, error) {
 }
 
 func getPackageExport(importPath string) (string, error) {
-	cmd := exec.Command("go", "list", "-export", "-f", "{{.Export}}", importPath)
+	dir := ""
 	if modDir := findModuleRoot(); modDir != "" {
-		cmd.Dir = modDir
+		dir = modDir
```
bitterpanda63: doesnt make sense to do separate definition in if file right if basic is just ""

cmd/zen-go/internal/importcfg.go
```diff
@@ -91,16 +93,30 @@ func createTempFile(objdir string) (*os.File, error) {
...
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(string(output)), nil
+
+	if len(pkgs) == 0 {
+		return "", errors.New("package was not resolved")
+	}
+
+	return pkgs[0].ExportFile, err
```
bitterpanda63, line 116: why do we only care about the first package's export file

tomaisthorpe: we're only asking for the export file of one package :)

bitterpanda63: aha, the import path, rightt, so it would be weird if there were two results then? would never happen I guess

bitterpanda63: Maybe worth a log if size > 1 or nah?

tomaisthorpe: I think probably not really necessary

bitterpanda63 (approved):

PR #270, opened 2026-01-14
Pass parent build flags to packages.Load

tomaisthorpe:
## Summary by Aikido
| ⚠️ Security Issues: 1 | 🔍 Quality Issues: 2 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Captured parent go command flags by adding buildflags.go implementation

**⚡ Enhancements**
* Passed parent build flags and environment into packages.Load calls

**🔧 Refactors**
* Simplified flag extraction, removed whitelist, and added overflow checks

<sup>[More info](https://app.aikido.dev/featurebranch/scan/74967554?groupId=6)</sup>


cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,171 @@
...
+
+	args, err := findParentGoProcess(goBinary)
+	if err != nil {
+		return nil, err
+	}
+
+	return extractBuildFlags(args), nil
+}
+
+func locateGoBinary() (string, error) {
```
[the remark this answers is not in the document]
tomaisthorpe: I don't think this would make the code meaningfully clearer and would just add indirection.

cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,171 @@
...
+		return false
+	}
+	return resolved == goBinary
+}
+
+// extractBuildFlags parses the parent go command's arguments and returns
+// the build flags. Skips -a (would force unnecessary rebuilds) and
+// -toolexec (we add our own).
+func extractBuildFlags(args []string) []string {
+	standalone := map[string]bool{
```
bitterpanda63: why is this allowlist instead of blocklist?

cmd/zen-go/internal/importcfg.go
```diff
@@ -95,6 +95,18 @@ func createTempFile(objdir string) (*os.File, error) {
 func getPackageExport(importPath string) (string, error) {
 	dir := findModuleRoot()
 
+	buildFlags, err := parentBuildFlags()
```
bitterpanda63, line 99: add comment that this is important for .. , I thought it was the hashes? or cache?

cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,171 @@
...
+import (
+	"errors"
+	"math"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/shirou/gopsutil/v4/process"
```
bitterpanda63, line 12: we use the process.NewProcess from this module, really no built-in alternative? adding a lot of deps with this

cmd/zen-go/internal/importcfg.go
```diff
@@ -95,6 +95,18 @@ func createTempFile(objdir string) (*os.File, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	// Add our toolexec so packages.Load uses the same instrumented cache
+	exe, err := os.Executable()
+	if err != nil {
+		return "", err
+	}
+	buildFlags = append(buildFlags, fmt.Sprintf("-toolexec=%s toolexec", exe))
```
bitterpanda63: what would happen with multiple -toolexec's? wouldn't this cause issues

cmd/zen-go/internal/buildflags.go
tomaisthorpe: not really, unfortunately there's a lot of platform specific stuff to do to get the parent process.

cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,128 @@
...
+		return "", err
+	}
+	path, err = filepath.Abs(path)
+	if err != nil {
+		return "", err
+	}
+	return filepath.EvalSymlinks(path)
+}
+
+func findParentGoProcess(goBinary string) ([]string, error) {
```
bitterpanda63: name is misleading, findParentGoProcessArguments right?

cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,128 @@
...
+	if pid < 0 || pid > math.MaxInt32 {
+		return nil, errors.New("PID out of int32 range")
+	}
+	pid32 := int32(pid)
+	proc, err := process.NewProcess(pid32)
+	if err != nil {
+		return nil, err
+	}
+
+	for {
```
bitterpanda63, line 68: this proc.Parent will give an error eventually? not fully understanding why this is all in a for loop, going through multiple parents or? can you add a comment maybe

bitterpanda63: assuming no risk of infinite recursion because either we will find the go binary command or the proc.Parent will be exhausted?

cmd/zen-go/internal/buildflags.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	for ; i < len(args); i++ {
+		arg := args[i]
+
+		if arg == "--" {
+			break
+		}
+
+		// Skip -a
```
bitterpanda63: useless comment, either remove or ideally specify further

cmd/zen-go/internal/buildflags.go
tomaisthorpe: yeah no risk, in a future PR I will ensure context gets passed down to these methods and we can use `ParentWithContext` too to ensure we respect any cancellations incase it does get stuck in a loop (but it shouldn't)

cmd/zen-go/internal/importcfg.go
```diff
@@ -95,6 +95,11 @@ func createTempFile(objdir string) (*os.File, error) {
 func getPackageExport(importPath string) (string, error) {
 	dir := findModuleRoot()
 
+	buildFlags, err := parentBuildFlags()
```
bitterpanda63, line 99:
```suggestion
    // prevent cache misses and fingerprint errors.
	buildFlags, err := parentBuildFlags()
```

bitterpanda63 (approved):

PR #274, opened 2026-01-16
Add attack wave samples

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Collected unique suspicious request samples per IP and exposed retrieval
* Included attack wave samples in cloud metadata and logged marshal errors
* Enabled attack wave QA test in workflow by unskipping it

**🔧 Refactors**
* Converted LRU cache to generics and parameterized value types

<sup>[More info](https://app.aikido.dev/featurebranch/scan/75274279?groupId=6)</sup>

### DIFF
--- a/internal/vulnerabilities/attackwave/lru_cache.go
+++ b/internal/vulnerabilities/attackwave/lru_cache.go
@@ -6,13 +6,13 @@ import (
 	"time"
 )
 
-type entry struct {
+type entry[T any] struct {
 	key    string
-	value  int
+	value  T
 	expiry time.Time
 }
 
-type lruCache struct {
+type lruCache[T any] struct {
 	mu       sync.Mutex
 	capacity int
 	ttl      time.Duration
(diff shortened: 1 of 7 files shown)


internal/vulnerabilities/attackwave/lru_cache.go
```diff
@@ -6,52 +6,53 @@ import (
...
 	}
 
-	ent := elem.Value.(*entry)
+	ent := elem.Value.(*entry[T])
 
 	// Check if expired
 	if time.Now().After(ent.expiry) {
 		c.removeElement(elem)
-		return 0, false
+		return zero, false
```
timokoessler, line 47: This change does not break existing usage?

tomaisthorpe: Nope! It will be doing the same as it was doing before. If T is an int, then it's still 0, if T is something else, it just ends up being whatever the zero value of that is. For example, if it's `*SomeStruct`, it will be a nil.

timokoessler (approved):

PR #277, opened 2026-01-19
Fix wrap rules versioned pkgs and add zen instrument for chi

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added zen instrumentation for go-chi with wrap rules and excludes

**⚡ Enhancements**
* Added exclude option to wrap rules and honored exclusions during instrumentation
* Handled module major-version suffixes when deriving import package names

<sup>[More info](https://app.aikido.dev/featurebranch/scan/76037387?groupId=6)</sup>

### DIFF
--- a/cmd/zen-go/internal/instrumentor.go
+++ b/cmd/zen-go/internal/instrumentor.go
@@ -70,7 +70,7 @@ func (i *Instrumentor) InstrumentFile(filename string, compilingPkg string) (Ins
 			parts := strings.Split(path, "/")
 			name = parts[len(parts)-1]
 			// Handle major version suffixes (e.g., /v5 in github.com/go-chi/chi/v5)
-			if len(parts) >= 2 && len(name) >= 2 && name[0] == 'v' && name[1] >= '0' && name[1] <= '9' {
+			if len(parts) >= 2 && isMajorVersion(name) {
 				name = parts[len(parts)-2]
 			}
 		}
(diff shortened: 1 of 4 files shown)


cmd/zen-go/internal/instrumentor.go
```diff
@@ -68,6 +69,10 @@ func (i *Instrumentor) InstrumentFile(filename string, compilingPkg string) (Ins
 		} else {
 			parts := strings.Split(path, "/")
 			name = parts[len(parts)-1]
+			// Handle major version suffixes (e.g., /v5 in github.com/go-chi/chi/v5)
+			if len(parts) >= 2 && len(name) >= 2 && name[0] == 'v' && name[1] >= '0' && name[1] <= '9' {
```
bitterpanda63: mm if thi works we don't need the exclude on /v5 no?

instrumentation/sources/go-chi/chi/zen.instrument.yml
```diff
@@ -0,0 +1,26 @@
+meta:
+  name: go-chi/chi
+  description: Chi is a lightweight, idiomatic and composable router for building Go HTTP services.
+
+rules:
+  - id: chi.NewMux
+    type: wrap
+    match: github.com/go-chi/chi/v5.NewMux
+    exclude:
+      - github.com/go-chi/chi/v5
```
bitterpanda63, line 10: this one

tomaisthorpe: That's not what it's doing, it's just excluding the package itself. The local package name is being used so we can find what the pkg is called within the code, for example, if they have something like an aliased import:

```go
import something "github.com/go-chi/chi/v5"

func main() {
  something.NewMux()
}
```

This is looking for `something`, so we can match with it.

bitterpanda63 (approved):

PR #311, opened 2026-02-06
Get directories of sub modules for loading rules

tomaisthorpe:
## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 2 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**⚡ Enhancements**
* Added functions to discover instrumentation directories across module and submodules
* Updated rules loader to skip subdirectories that are separate Go modules
* Aggregated and loaded instrumentation rules from multiple discovered instrumentation directories

<sup>[More info](https://app.aikido.dev/featurebranch/scan/82007139?groupId=6)</sup>

### DIFF
--- a/cmd/zen-go/internal/rules/rules.go
+++ b/cmd/zen-go/internal/rules/rules.go
@@ -75,7 +75,9 @@ type InjectDeclRule struct {
 	DeclTemplate string   // The declaration to inject
 }
 
-// LoadRulesFromDir loads all zen.instrument.yml files from a directory tree
+// LoadRulesFromDir loads all zen.instrument.yml files from a directory tree.
+// Subdirectories that contain their own go.mod are skipped, since those are
+// separate Go modules whose rules will be discovered independently.
 func LoadRulesFromDir(dir string) (*InstrumentationRules, error) {
 	result := &InstrumentationRules{}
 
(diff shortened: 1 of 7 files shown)


cmd/zen-go/internal/rules/rules.go
```diff
@@ -84,7 +86,18 @@ func LoadRulesFromDir(dir string) (*InstrumentationRules, error) {
 			return err
 		}
 
-		if d.IsDir() || !strings.HasSuffix(d.Name(), "zen.instrument.yml") {
+		if d.IsDir() {
+			// Skip subdirectories that are separate Go modules (have their own go.mod).
```
bitterpanda63, line 90: Could we not filter the root dir for sub moudles in paths.go and have a clearer separation of concerns?

tomaisthorpe: I think it currently has clear separation of concerns. The submodule skipping feels like a directory-walking concern (how to load the rules), and not a path-discovery concern (what to load). With the current structure, `LoadRulesFromDir` should be correct regardless of how the caller filters the inputs. I think moving that filtering to the paths package would result in more coupling.

bitterpanda63: kk, agreed

bitterpanda63 (approved):

PR #338, opened 2026-02-12
Strip port from user IP in net/http source

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Stripped port from request RemoteAddr and parsed IPv4/IPv6

<sup>[More info](https://app.aikido.dev/featurebranch/scan/83043893?groupId=6)</sup>

### DIFF
--- a/instrumentation/sources/net/http/middleware.go
+++ b/instrumentation/sources/net/http/middleware.go
@@ -1,8 +1,10 @@
 package http
 
 import (
+	"log/slog"
 	"mime/multipart"
 	"net/http"
+	"net/netip"
 	"strings"
 	"sync/atomic"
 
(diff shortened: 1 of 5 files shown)


instrumentation/sources/net/http/middleware.go
```diff
@@ -22,7 +24,13 @@ func Middleware(orig func(w http.ResponseWriter, r *http.Request)) func(w http.R
 			return
 		}
 
-		ip := r.RemoteAddr
+		var ip string
+		addrPort, err := netip.ParseAddrPort(r.RemoteAddr)
```
timokoessler, line 28: RemoteAddr always includes a port?

bitterpanda63: think the idea is to strip port if it is there, but @tomaisthorpe maybe the log should have an update, because right now the "could not parse ip" might be a bit misleading. I'm assuming this is fine if no port provided then it gives error and we just keep the original ip string

tomaisthorpe: Yep, always includes the port, can see it here:
https://cs.opensource.google/go/go/+/refs/tags/go1.26.0:src/net/http/request.go;l=292

I'm not sure what else the log message should be updated to? 
Also, we do need to improve this a little to handle forwarded IPs, etc.

bitterpanda63 (approved):

PR #370, opened 2026-03-12
Add hooks package for bridging to agent

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added hooks package to bridge sinks to agent runtime
* Added operation.Kind type and constants for operation kinds

**⚡ Enhancements**
* Registered agent runtime and implemented hooks.Runtime bridge methods

**🔧 Refactors**
* Replaced direct agent calls in sinks with hooks invocations
* Converted stats operation kinds to use new operation.Kind type

<sup>[More info](https://app.aikido.dev/featurebranch/scan/92584820?groupId=6)</sup>

### DIFF
--- a/instrumentation/hooks/hooks.go
+++ b/instrumentation/hooks/hooks.go
@@ -1,0 +1,47 @@
+package hooks
+
+// OperationKind identifies the type of operation being tracked.
+type OperationKind string
+
+const (
+	OperationKindSQL          OperationKind = "sql_op"
+	OperationKindNoSQL        OperationKind = "nosql_op"
+	OperationKindOutgoingHTTP OperationKind = "outgoing_http_op"
+	OperationKindFileSystem   OperationKind = "fs_op"
+	OperationKindExec         OperationKind = "exec_op"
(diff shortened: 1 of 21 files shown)


instrumentation/hooks/hooks.go
```diff
@@ -0,0 +1,47 @@
+package hooks
+
+// OperationKind identifies the type of operation being tracked.
+type OperationKind string
+
+const (
+	OperationKindSQL          OperationKind = "sql_op"
```
bitterpanda63: a tiny bit dirty that this is put in the hooks.go, because the constants here have ofc nothiing to do with the hooks logic but with one of the hooks, but not a big deal

bitterpanda63: would avoid a double definition maybe though

tomaisthorpe: added an operation package to share this kind value.

bitterpanda63 (approved):

PR #374, opened 2026-03-17
Add stored SSRF checks

tomaisthorpe:
## Summary by Aikido
|  Security Issues: 0 | 🔍 Quality Issues: 3 |  Resolved Issues: 0 |
| :--- | :--- | :--- |

**🚀 New Features**
* Added stored SSRF detection and reporting functionality across modules.

**⚡ Enhancements**
* Integrated stored SSRF checks into HTTP dialer to block connections.
* Propagated bypassed request flag via GLS and ensured context propagation.
* Added force-protection-off check and enabled stored SSRF QA tests.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/97424880?groupId=6)</sup>

### DIFF
--- a/vulnerabilities/attack.go
+++ b/vulnerabilities/attack.go
@@ -24,6 +24,8 @@ const (
 	KindShellInjection AttackKind = "shell_injection"
 	// KindSSRF indicates a server-side request forgery attack was detected.
 	KindSSRF AttackKind = "ssrf"
+	// KindStoredSSRF indicates a stored server-side request forgery attack was detected.
+	KindStoredSSRF AttackKind = "stored_ssrf"
 )
 
 func getDisplayNameForAttackKind(kind AttackKind) string {
@@ -36,6 +38,8 @@ func getDisplayNameForAttackKind(kind AttackKind) string {
 		return "a shell injection"
 	case KindSSRF:
 		return "a server-side request forgery"
+	case KindStoredSSRF:
+		return "a stored server-side request forgery"
 	default:
 		return "unknown attack type"
 	}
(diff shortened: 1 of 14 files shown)


vulnerabilities/ssrf/imds.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/AikidoSec/firewall-go/internal/agent/ipaddr"
+)
+
+var imdsIPList = ipaddr.BuildMatchList("imds", "IMDS IP addresses", []string{
+	"169.254.169.254",
+	"100.100.100.200",
+	"fd00:ec2::254",
+	"::ffff:169.254.169.254",
```
bitterpanda63: let's make sure the IP parsers just has this logic no? prone to fail if someone quickly adds only the ipv4 ip

vulnerabilities/ssrf/stored_ssrf.go
```diff
@@ -0,0 +1,22 @@
...
+type StoredSSRFResult struct {
+	Hostname  string
+	PrivateIP string
+}
+
+// CheckStoredSSRF checks if a hostname resolves to an IMDS IP address,
+// indicating a potential stored SSRF attack (e.g., DNS spoofing or /etc/hosts poisoning).
+// Returns nil if no attack is detected.
+func CheckStoredSSRF(hostname string, resolvedIPs []string) *StoredSSRFResult {
+	imdsIP := resolvesToIMDSIP(resolvedIPs, hostname)
```
bitterpanda63: we are certain a hostname can never be an IP right? - this is the case for some of our other langs and if that can happen basically would block all imds apart from mb the google one

vulnerabilities/attack.go
```diff
@@ -166,6 +170,63 @@ func storeDeferredAttack(ctx context.Context, res *interceptorResult) error {
...
+		}
+		user = reqCtx.GetUser()
+	}
+
+	attack := &agent.DetectedAttack{
+		Request: reqInfo,
+		Attack: aikido_types.AttackDetails{
+			Kind:      string(KindStoredSSRF),
+			Operation: operation,
+			Module:    "Module",
```
bitterpanda63: can we not just leave this empty, having Module as the empty placeholder could cause edge cases in the future right

vulnerabilities/ssrf/imds.go
tomaisthorpe: Just checked, not needed at all, gone!

vulnerabilities/attack.go
tomaisthorpe: Passing a module name from the sink now!

vulnerabilities/ssrf/stored_ssrf_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCheckStoredSSRF(t *testing.T) {
+	t.Run("returns result with hostname and privateIP", func(t *testing.T) {
```
bitterpanda63, line 11: lets add a test case with ipv6 mapped ips here as well

tomaisthorpe: added!

vulnerabilities/ssrf/stored_ssrf.go
tomaisthorpe: this is checked within resolvesToIMDSIP itself, but i've also added some tests to check this.

vulnerabilities/ssrf/stored_ssrf.go
```diff
@@ -0,0 +1,32 @@
...
+
+// CheckStoredSSRF checks if a hostname resolves to an IMDS IP address,
+// indicating a potential stored SSRF attack (e.g., DNS spoofing or /etc/hosts poisoning).
+// Returns nil if no attack is detected.
+func CheckStoredSSRF(hostname string, resolvedIPs []string) *StoredSSRFResult {
+	if isTrustedHostname(hostname) {
+		return nil
+	}
+
+	if slices.Contains(resolvedIPs, hostname) {
```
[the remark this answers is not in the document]
tomaisthorpe: Resolved IPs will be very small, doesn't have significant impact.

vulnerabilities/ssrf/stored_ssrf.go
```diff
@@ -0,0 +1,32 @@
...
+		return nil
```
bitterpanda63, line 21:
```suggestion
        // if the hostname itself is an IP then we don't want to block this, as it's just an IMDS request, not stored ssrf.
		return nil
```

bitterpanda63 (approved):

PR #382, opened 2026-03-26
Skip outbound blocking for bypassed IPs

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Skipped outbound blocking for requests flagged as bypassed while reporting.

**🔧 Refactors**
* Added internal request import to support bypassed-IP checking functionality.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/97030640?groupId=6)</sup>

### DIFF
--- a/instrumentation/sinks/net/http/examine.go
+++ b/instrumentation/sinks/net/http/examine.go
@@ -6,11 +6,16 @@ import (
 
 	"github.com/AikidoSec/firewall-go/instrumentation/hooks"
 	"github.com/AikidoSec/firewall-go/instrumentation/operation"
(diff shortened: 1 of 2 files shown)


instrumentation/sinks/net/http/examine.go
timokoessler, line 28: We still want to report them to core? At least we are doing so in the Node.js agent I think.

tomaisthorpe: yep, you're correct, updated!

timokoessler (approved):

PR #388, opened 2026-03-27
Block SQL queries that fail tokenization

hansott:
Our SQL injection detection tokenizes queries to check them. If the tokenizer can't parse a query, we skip it and the query goes through. Some databases still execute partially valid queries though: ClickHouse ignores junk after ; and SQLite runs everything before an unclosed /*.

Now when user input shows up in a query we can't tokenize, we treat it as an attack. On by default, opt out with AIKIDO_BLOCK_INVALID_SQL=false.

## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Blocked queries that failed tokenization by default, opt-out enabled

**🔧 Refactors**
* Replaced boolean detector with integer status codes and constants
* Updated scanFunction to handle detector result codes and imports

<sup>[More info](https://app.aikido.dev/featurebranch/scan/97404896?groupId=6)</sup>

### DIFF
--- a/vulnerabilities/sqlinjection/vulnerability.go
+++ b/vulnerabilities/sqlinjection/vulnerability.go
@@ -39,7 +39,7 @@ func scanFunction(userInput string, args *ScanArgs) (*vulnerabilities.ScanResult
 
 	result := detectSQLInjection(statement, userInput, dialect)
 
-	if result == 1 {
+	if result == sqlInjectionDetected {
 		return &vulnerabilities.ScanResult{
 			DetectedAttack: true,
 			Metadata: map[string]string{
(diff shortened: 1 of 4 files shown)


vulnerabilities/sqlinjection/vulnerability.go
```diff
@@ -36,5 +49,16 @@ func scanFunction(userInput string, args *ScanArgs) (*vulnerabilities.ScanResult
 		}, nil
 	}
 
+	if result == 3 && shouldBlockInvalidSqlQueries() {
```
tomaisthorpe: Can these magic numbers be constants to make it clear what they represent?

hansott: https://github.com/AikidoSec/firewall-go/pull/388/commits/1904081234e6177eef0b576e127a8e17da80d5f5

tomaisthorpe (approved):

PR #441, opened 2026-05-15
Track custom events

tomaisthorpe:


internal/agent/cloud/event_custom_event.go
```diff
@@ -0,0 +1,15 @@
+package cloud
+
+type CustomEvent struct {
```
timokoessler, line 3: Todo: This event needs to be updated to follow new spec:  https://github.com/AikidoSec/zen-specs/blob/main/features/custom_event_tracking.md

PR #444, opened 2026-05-20
Add automatic goroutine propagation

tomaisthorpe:
## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added runtime instrumentation to propagate request context across goroutines.

**⚡ Enhancements**
* Introduced testutil fallback GLS using parsed goroutine IDs for tests.
* Updated bench script and instrumentation imports to support runtime benchmarks.

**🔧 Refactors**
* Refactored internal request GLS to register and use injected accessors.

<sup>[More info](https://app.aikido.dev/featurebranch/scan/132500203?groupId=6)</sup>

### DIFF
--- a/instrumentation/runtime/_runtime/aikido_gls.go
+++ b/instrumentation/runtime/_runtime/aikido_gls.go
@@ -1,0 +1,28 @@
+// This file is added to the standard library "runtime" package by the
+// zen-go toolexec via the add-file rule in zen.instrument.yml. It is NOT
+// compiled as part of any normal Go build; the leading underscore in the
+// parent directory keeps the Go toolchain from discovering it.
+//
+// Once injected, it provides accessors for the aikido_request_context field
+// that the add-field rule adds to runtime.g, and it registers those accessors
+// with firewall-go's internal/request package so the rest of the firewall can
+// read and write the per-goroutine value without using a third-party GLS.
+
+package runtime
+
+import _ "unsafe"
+
+//go:linkname aikidoRegisterGLS github.com/AikidoSec/firewall-go/internal/request.RegisterGLS
+func aikidoRegisterGLS(get func() interface{}, set func(interface{}))
+
+func aikidoGLSGet() interface{} {
+	return getg().m.curg.aikido_request_context
+}
+
+func aikidoGLSSet(v interface{}) {
+	getg().m.curg.aikido_request_context = v
+}
+
+func init() {
+	aikidoRegisterGLS(aikidoGLSGet, aikidoGLSSet)
+}
--- a/instrumentation/runtime/_runtime/aikido_gls.go
+++ b/instrumentation/runtime/_runtime/aikido_gls.go
@@ -15,6 +15,9 @@ import _ "unsafe"
 //go:linkname aikidoRegisterGLS github.com/AikidoSec/firewall-go/internal/request.RegisterGLS
 func aikidoRegisterGLS(get func() interface{}, set func(interface{}))
 
+// getg() returns the current g (goroutine struct). m is the OS thread it's
+// running on; curg is the user goroutine scheduled on that thread. We use curg
+// because getg() can return a system goroutine when called from runtime internals.
 func aikidoGLSGet() interface{} {
 	return getg().m.curg.aikido_request_context
 }
--- a/instrumentation/runtime/zen.instrument.yml
+++ b/instrumentation/runtime/zen.instrument.yml
@@ -1,0 +1,29 @@
+meta:
+  name: runtime
+  description: Automatic propagation of Aikido request context across goroutines.
+
+rules:
+  # Add a field to runtime.g that stores the per-goroutine request state.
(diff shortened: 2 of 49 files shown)


instrumentation/runtime/_runtime/aikido_gls.go
```diff
@@ -0,0 +1,28 @@
...
+
+package runtime
+
+import _ "unsafe"
+
+//go:linkname aikidoRegisterGLS github.com/AikidoSec/firewall-go/internal/request.RegisterGLS
+func aikidoRegisterGLS(get func() interface{}, set func(interface{}))
+
+func aikidoGLSGet() interface{} {
+	return getg().m.curg.aikido_request_context
```
timokoessler, line 22: `getg().m` could be `nil` in some edge cases?

internal/testutil/gls.go
```diff
@@ -0,0 +1,46 @@
...
+	})
+}
+
+func goid() int64 {
+	var buf [64]byte
+	n := runtime.Stack(buf[:], false)
+	s := strings.TrimPrefix(string(buf[:n]), "goroutine ")
+	idx := strings.IndexByte(s, ' ')
+	if idx < 0 {
+		return 0
```
timokoessler: If this happens it will be hard to debug and result in strange effects?

timokoessler (commented):
Add some e2e tests?

instrumentation/runtime/_runtime/aikido_gls.go
tomaisthorpe: `getg()` returns the currently running goroutine, which is _always_ bound to it's `m`, which is the OS thread. 

There's a little info here:
https://github.com/golang/go/blob/master/src/runtime/HACKING.md#getg-and-getgmcurg

internal/testutil/gls.go
tomaisthorpe: agreed, if it fails now will panic instead.

tomaisthorpe:
> Add some e2e tests?

@timokoessler, this was already being exercised in the new OS integration test, and I have a change ready to go on the demo app once this is in a release. However, just to add more coverage, I've also adjusted two of the sample apps to read from a goroutine, so we'll also see it exercised in the e2e tests.

timokoessler (approved):
Might be good to have another pair of eyes tho.

instrumentation/runtime/_runtime/aikido_gls.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func aikidoGLSSet(v interface{}) {
+	getg().m.curg.aikido_request_context = v
```
bitterpanda63, line 26: whats "curg"?

tomaisthorpe: g = goroutine
curg = current goroutine

https://github.com/golang/go/blob/master/src/runtime/HACKING.md#getg-and-getgmcurg

instrumentation/runtime/zen.instrument.yml
```diff
@@ -0,0 +1,30 @@
+meta:
+  name: runtime
+  description: Automatic propagation of Aikido request context across goroutines.
+  min-zen-go-version: "1.0.3"
+
+rules:
+  # Add a field to runtime.g that stores the per-goroutine request state.
```
bitterpanda63, line 7:
```suggestion
  # Add a field to runtime.g that stores the per-goroutine request state.
  # https://github.com/golang/go/blob/master/src/runtime/HACKING.md#getg-and-getgmcurg
```

internal/request/gls.go
```diff
@@ -2,52 +2,72 @@ package request
...
+type glsState struct {
+	ctx      *Context
+	bypassed bool
+}
+
+func getGLS() *glsState {
+	if glsGet == nil {
+		return nil
+	}
+	v := glsGet()
```
bitterpanda63: could we use a tiny bit more descriptive names than `v` and `s` ?

internal/request/gls.go
```diff
@@ -2,52 +2,72 @@ package request
...
 	}
 
-	if IsBypassed(ctx) {
-		glsManager.SetValues(gls.Values{
-			glsBypassedKey: true,
-		}, fn)
+	var state *glsState
+	if reqCtx := GetContext(ctx); reqCtx != nil {
+		state = &glsState{ctx: reqCtx}
+	} else if IsBypassed(ctx) {
```
bitterpanda63: is now an else if, this didnt use to be the case right?

tomaisthorpe: Yep! it's just how we build up the state has changed a little. We don't have the early returns anymore with the `SetValues`, remember before it was relying on being within the same callstack, so we had things like this.

tomaisthorpe: changed my mind though, going to refactor it slightly.

bitterpanda63 (approved):

PR #445, opened 2026-05-20
Add test to check for goroutine propagation for `os`

tomaisthorpe:


tomaisthorpe:
This was only proof for #444, closing.

PR #479, opened 2026-06-05
fix: Preserve empty properties in API schema

timokoessler:
## Summary by Aikido
| :--- | :--- | :--- |

**⚡ Enhancements**
* Preserved empty schema fields by removing omitempty from DataSchema

<sup>[More info](https://app.aikido.dev/featurebranch/scan/126909162?groupId=6)</sup>

### DIFF
--- a/internal/agent/aikido_types/api_spec.go
+++ b/internal/agent/aikido_types/api_spec.go
@@ -10,9 +10,9 @@ type APIAuthType struct {
 
 type DataSchema struct {
 	Type       []string               `json:"type"`
-	Properties map[string]*DataSchema `json:"properties,omitempty"`
-	Items      *DataSchema            `json:"items,omitempty"`
-	Optional   bool                   `json:"optional,omitempty"`
+	Properties map[string]*DataSchema `json:"properties"`
+	Items      *DataSchema            `json:"items"`
+	Optional   bool                   `json:"optional"`
 }
 
 type APIBodyInfo struct {
--- a/internal/agent/apidiscovery/merge_data_schemas_test.go
+++ b/internal/agent/apidiscovery/merge_data_schemas_test.go
@@ -1,11 +1,13 @@
 package apidiscovery
 
 import (
+	"encoding/json"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


internal/agent/aikido_types/api_spec.go
```diff
@@ -10,9 +10,9 @@ type APIAuthType struct {
 
 type DataSchema struct {
 	Type       []string               `json:"type"`
-	Properties map[string]*DataSchema `json:"properties,omitempty"`
-	Items      *DataSchema            `json:"items,omitempty"`
-	Optional   bool                   `json:"optional,omitempty"`
+	Properties map[string]*DataSchema `json:"properties"`
+	Items      *DataSchema            `json:"items"`
+	Optional   bool                   `json:"optional"`
```
tomaisthorpe, line 15: I don't think `omitempty` should have been removed from these two. 

It will result in `"items": null` for every item that isn't an array, and will always include `"optional": false` when not necessary.

internal/agent/apidiscovery/merge_data_schemas_test.go
```diff
@@ -216,3 +218,32 @@ func TestMergeTypes(t *testing.T) {
...
+		data, err := json.Marshal(empty)
+		require.NoError(t, err)
+		var restoredEmpty aikido_types.DataSchema
+		require.NoError(t, json.Unmarshal(data, &restoredEmpty))
+
+		result := MergeDataSchemas(withProps, &restoredEmpty)
+		assert.NotNil(t, result.Properties)
+		assert.Contains(t, result.Properties, "key")
+	})
+}
```
tomaisthorpe: I feel like I'm not sure what this test is doing that validates this issue. The merging of nil/empty values is quite different to JSON marshaling/unmarshaling.

internal/agent/aikido_types/api_spec.go
timokoessler: ~Changing to omitzero~

timokoessler: Actually we send items: null in other agents

tomaisthorpe (approved):

PR #490, opened 2026-06-24
Add 'go' command to zen-go CLI for 'go tool' support

tomaisthorpe:
Allows for running `zen-go` with `go tool zen-go go build .` instead of having to add the toolexec, and allows us to pin the tool version in go.mod.

I've updated the sample apps to already use the `zen-go go build .` approach.

## Summary by Aikido
| :--- | :--- | :--- |

**🚀 New Features**
* Added 'go' command to run go tool with Zen instrumentation

**⚡ Enhancements**
* Enhanced quoting logic to handle single and double quotes in paths

<sup>[More info](https://app.aikido.dev/featurebranch/scan/137268800?groupId=6)</sup>

### DIFF
--- a/cmd/zen-go/cmd_go.go
+++ b/cmd/zen-go/cmd_go.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+)
+
+var toolexecSubcommands = map[string]bool{
+	"build":   true,
+	"install": true,
+	"run":     true,
+	"test":    true,
+	"vet":     true,
+}
+
+func buildGoArgs(self string, args []string) []string {
+	if !toolexecSubcommands[args[0]] {
+		return args
+	}
+
+	toolexec := fmt.Sprintf("-toolexec=%s toolexec", self)
+
+	goArgs := make([]string, 0, len(args)+1)
+	goArgs = append(goArgs, args[0], toolexec)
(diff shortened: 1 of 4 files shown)


cmd/zen-go/cmd_go.go
```diff
@@ -0,0 +1,53 @@
...
+	"test":    true,
+	"vet":     true,
+}
+
+func buildGoArgs(self string, args []string) []string {
+	if !toolexecSubcommands[args[0]] {
+		return args
+	}
+
+	toolexec := fmt.Sprintf("-toolexec=%s toolexec", self)
```
timokoessler: What if the path to the zen go binary contains spaces?

cmd/zen-go/cmd_go_test.go
timokoessler, line 1: Any tests that are actually executing the cli btw?

tomaisthorpe: Yep, as I've changed the sample apps to use `zen-go go build ...`, it means they are using this method for the e2e tests.

cmd/zen-go/cmd_go.go
tomaisthorpe: Think you're right! go [this](https://go.dev/src/cmd/internal/quoted/quoted.go) to split the strings, so we can just quote them in either single or double quotes and then we're good. Updated it.

timokoessler (approved):

PR #534, opened 2026-07-27
Fix package level net/http Handle/HandlerFunc not being instrumented

tomaisthorpe:
The package level methods call internal methods instead of the public Handle/HandlerFunc on the default serve mux:
https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/net/http/server.go;l=2862

### DIFF
--- a/instrumentation/sources/net/http/zen.instrument.yml
+++ b/instrumentation/sources/net/http/zen.instrument.yml
@@ -40,3 +40,19 @@ rules:
     function: HandleFunc
     template: |
       {{ .Function.Argument 1 }} = __aikido_http_middleware({{ .Function.Argument 1 }})
+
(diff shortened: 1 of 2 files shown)


instrumentation/sources/net/http/zen.instrument.yml
```diff
@@ -40,3 +40,19 @@ rules:
     function: HandleFunc
     template: |
       {{ .Function.Argument 1 }} = __aikido_http_middleware({{ .Function.Argument 1 }})
+
```
timokoessler, line 43: Can't this change lead to double instrumentation?

tomaisthorpe: That's why I didn't do this originally :)

But it turns out those package levels methods don't call the exported methods on the mux receiver, so it has to be instrumented separately. You can check the Go source link in the PR description to see it.

timokoessler (approved):

PR #537, opened 2026-07-30
Add support for fiber/v2

tomaisthorpe:

PR #539, opened 2026-07-30
Decouple request context from *http.Request

tomaisthorpe:
Split from #537 

To be able to support frameworks that don't use `net/http`, we need to decouple our request context from it. Now passing data directly instead of processing within the request context.

### DIFF
--- a/instrumentation/http/request_helpers.go
+++ b/instrumentation/http/request_helpers.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
+	"github.com/AikidoSec/firewall-go/internal/request"
 )
 
 func FullURL(r *http.Request) string {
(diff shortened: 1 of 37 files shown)


instrumentation/http/request_helpers.go
```diff
@@ -0,0 +1,50 @@
...
+	if r.TLS != nil {
+		scheme = "https"
+	}
+	return fmt.Sprintf("%s://%s%s", scheme, r.Host, r.URL.RequestURI())
+}
+
+func HeadersToMap(headers http.Header) map[string][]string {
+	result := make(map[string][]string, len(headers))
+	for key, values := range headers {
+		if strings.ToLower(key) == "cookie" {
```
timokoessler, line 22: Does this allow any bypasses? E.g. if the whole header is used in an injection (unparsed)? Unlikely, I know, but we are also adding the full encoded JWT to the context because of a past bug bounty report.

timokoessler (approved):

instrumentation/http/request_helpers.go
tomaisthorpe: Are you thinking an app might use the cookie header somewhere directly?

tomaisthorpe: Will fix in next PR 🫡

